PR #2, opened 2017-05-12
integration tests initial setup

keymone:
- makefile
- dumb itest
- dockerfiles cleanup


tests/integration/mesos/features/sync.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Mesos Sync runner
```
sjaensch: Any reason we're using bdd style tests here? Unless it provides a compelling advantage I'd vote for sticking to one style of writing tests.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+itest:
+	cd tests/integration && \
+	docker-compose -f cluster/docker-compose.yaml down && \
+	docker-compose -f cluster/docker-compose.yaml pull && \
+	docker-compose -f cluster/docker-compose.yaml build && \
+	docker-compose -f cluster/docker-compose.yaml up -d zookeeper mesosmaster mesosslave && \
+	docker-compose -f cluster/docker-compose.yaml scale mesosslave=2 && \
+	docker-compose -f cluster/docker-compose.yaml run playground /src/itest.sh
+
+.taskproc/bin/activate:
```
sjaensch: How about, instead of just creating a hidden virtualenv for installing tox, we create one that will function as the development environment as well, i.e. where we install all of task_processing's runtime and development dependencies into?

Makefile
```diff
@@ -0,0 +1,27 @@
+.PHONY: all docs test itest
+
+docs: .taskproc/bin/activate
+	.taskproc/bin/tox -e docs
+
+test: .taskproc/bin/activate
```
sjaensch: You should also make sure pre-commit is installed, doing it as part of `make test` is a good idea - either here or from within tox.

tests/integration/mesos/_test_mesos.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
```
sjaensch: Once you got pre-commit set up please run it on all files. These imports are not necessary for Python 3.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+	.taskproc/bin/tox
+
+itest:
+	cd tests/integration && \
+	docker-compose -f cluster/docker-compose.yaml down && \
```
sjaensch: `/bin/sh: docker-compose: command not found`
How about doing all of this within tox? That way all knowledge on how to run tests lives there. As you notice you'll probably want to start a requirements-dev.txt as well.

tests/integration/mesos/features/sync.feature
keymone: for now this is sort of abandoned, but in general i don't see the problem with unit tests being unit tests and integration being behavior-driven. ideally i don't want to even touch our python libs in integration tests?

Makefile
keymone: we will have it setup in repo's git hooks

jolynch:
I'm going to merge this and we are going to iterate.

PR #6, opened 2017-05-13
makefile stuff

keymone:
- smaller makefile
- .tox/dev env instead of .taskproc
- move integration tests into tox

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,3 +18,16 @@ deps = sphinx
 changedir = docs
 commands =
     sphinx-build -b html -d build/doctrees source build/html
+
+[testenv:integration]
+deps =
+  docker-compose
+commands =
+  docker-compose -f examples/cluster/docker-compose.yaml down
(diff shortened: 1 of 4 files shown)


keymone:
this is supposed to fail, we neither have coverage yet nor docs nor mesos.native available

tox.ini
```diff
@@ -7,14 +7,27 @@ skipsdist=True
...
-	pytest -v {posargs:tests}
+  pytest -v {posargs:tests}
 
 [testenv:docs]
 deps = sphinx
 changedir = docs
 commands =
     sphinx-build -b html -d build/doctrees source build/html
+
+[testenv:integration]
```
jolynch: Can we just noop the integration target so that travis still passes until we do?

We can have a local enabled target for testing of course.

jolynch:
Mind a rebase?

tox.ini
keymone: makes sense

keymone:
@jolynch rebased and commented out failing tests

PR #8, opened 2017-05-15
Simplify mesos scheduler and remove references to paasta.

sagar8192:
I have refactored the mesos scheduler to remove references to paasta and some other unnecessary stuff. @Rob-Johnson @keymone @jolynch please take a look :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,6 @@
         files: ^tests/.+\.py$
         language_version: python3.6
     -   id: flake8
-        args: [--max-line-length=120]
         language_version: python3.6
     -   id: requirements-txt-fixer
         language_version: python3.6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,6 +18,7 @@
         files: ^tests/.+\.py$
         language_version: python3.6
     -   id: flake8
+        args: [--max-line-length=120]
         language_version: python3.6
     -   id: requirements-txt-fixer
         language_version: python3.6
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -20,10 +20,16 @@ logging.basicConfig(format=FORMAT, level=LEVEL)
 log = logging.getLogger(__name__)
 
 
+# TODO: Move this to utils
+def start_new_thread(func):
+    t = threading.Thread(target=func, args=())
+    t.start()
+
+
 class ExecutionFramework(mesos.interface.Scheduler):
     def __init__(
         self,
-        name="remote_run",
+        name,
         task_staging_timeout_s=240,
         pool=None,
         max_task_queue_size=1000,
@@ -52,13 +58,8 @@ class ExecutionFramework(mesos.interface.Scheduler):
         self.blacklisted_slaves = {}
         self.task_metadata = {}
 
-        self.start_new_thread(self.kill_tasks_stuck_in_staging)
-        self.start_new_thread(self.unblacklist_slaves)
-
-    def start_new_thread(self, func):
-        t = threading.Thread(target=func, args=())
-        # t.set_daemon = True
-        t.start()
+        start_new_thread(self.kill_tasks_stuck_in_staging)
+        start_new_thread(self.unblacklist_slaves)
 
     def kill_tasks_stuck_in_staging(self):
         while True:
@@ -71,7 +72,8 @@ class ExecutionFramework(mesos.interface.Scheduler):
                     log.warning('Killing stuck task {id}'.format(id=task_id))
                     self.kill_task(task_id)
                     self.blacklist_slave(
-                        self.task_metadata[task_id]['slave_id'])
+                        self.task_metadata[task_id]['slave_id']
+                    )
             time.sleep(10)
 
     def offer_matches_pool(self, offer):
@@ -96,8 +98,7 @@ class ExecutionFramework(mesos.interface.Scheduler):
         log.info('Blacklisting slave: {id} for {secs} seconds.'.format(
             id=slave_id,
             secs=self.slave_blacklist_timeout_s
-        )
-        )
+        ))
         self.blacklisted_slaves[slave_id] = time.time()
 
     def unblacklist_slaves(self):
@@ -170,7 +171,8 @@ class ExecutionFramework(mesos.interface.Scheduler):
                 available_ports = self.get_available_ports(resource)
 
         log.info(
-            "Received offer {id} with cpus: {cpu}, mem: {mem}, disk: {disk}".format(
+            "Received offer {id} with cpus: {cpu}, mem: {mem}, \
+            disk: {disk}".format(
                 id=offer.id.value,
                 cpu=remaining_cpus,
                 mem=remaining_mem,
@@ -187,8 +189,13 @@ class ExecutionFramework(mesos.interface.Scheduler):
                  remaining_mem >= task.mem and
                  remaining_disk >= task.disk)):
                 # This offer is sufficient for us to launch task
-                tasks_to_launch.append(self.create_new_docker_task(
-                    offer, task, available_ports))
+                tasks_to_launch.append(
+                    self.create_new_docker_task(
+                        offer,
+                        task,
+                        available_ports
+                    )
+                )
 
                 # Deduct the resources taken by this task from the total
                 # available resources.
@@ -291,11 +298,13 @@ class ExecutionFramework(mesos.interface.Scheduler):
 
     def registered(self, driver, frameworkId, masterInfo):
         log.info("Registered with framework ID {id}".format(
-            id=frameworkId.value))
+            id=frameworkId.value
+        ))
 
     def reregistered(self, driver, frameworkId, masterInfo):
         log.warning("Registered with framework ID {id}".format(
-            id=frameworkId.value))
+            id=frameworkId.value
+        ))
 
     def resourceOffers(self, driver, offers):
         if self.driver is None:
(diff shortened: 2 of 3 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +18,103 @@
...
 class ExecutionFramework(mesos.interface.Scheduler):
     def __init__(
         self,
-        name,
-        staging_timeout,
-        reconcile_backoff=1,
-        reconcile_start_time=float('inf'),
-        dry_run=False,
+        name="remote_run",
+        task_stagging_timeout_s=240,
```
keymone: typo

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +18,103 @@
...
+
+        self.start_new_thread(self.kill_tasks_stuck_in_stagging)
+        self.start_new_thread(self.unblacklist_slaves)
+
+    def start_new_thread(self, func):
+        t = threading.Thread(target=func, args=())
+        t.start()
+
+    def kill_tasks_stuck_in_stagging(self):
+        import time
```
keymone: move the import. pre-commit should have caught this, did you rebase?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +267,95 @@ def create_new_docker_task(self, offer, task_config):
...
+                    id=offer.id.value,
+                    slave=offer.slave_id.value
+                )
+            )
+            driver.launchTasks(offer.id, tasks_to_launch)
 
-        driver.acknowledgeStatusUpdate(update)
+            for task in tasks_to_launch:
+                self.tasks_launched[task.task_id.value] = time.time()
+                self.task_to_slave_id_mappings[task.task_id.value] = task.slave_id.value
```
keymone: i'd rather store a single compound value for each task instead of having to maintain multiple dictionaries

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +267,95 @@ def create_new_docker_task(self, offer, task_config):
...
 
-        self.reconcile_start_time = time.time()
-        driver.reconcileTasks([])
+    def reregistered(self, driver, frameworkId, masterInfo):
+        log.warning("Registered with framework ID {id}".format(id=frameworkId.value))
 
     def resourceOffers(self, driver, offers):
-        if self.frozen:
-            return
+        if self.driver is None:
```
keymone: i understand that in current implementation we need driver to be accessible from inside the framework, but i'm worried this will lead to bugs when driver object becomes stale? not sure how to go about solving this. reassign self.driver every time mesos calls framework methods?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +267,95 @@ def create_new_docker_task(self, offer, task_config):
...
-                if offer.slave_id.value in self.blacklisted_slaves:
-                    log.critical(
-                        'Ignoring offer {0} from blacklisted slave {1}'.format(
-                            offer.id.value, offer.slave_id.value
-                        )
+        for offer in offers:
+            if self.task_queue.empty():
+                log.info("Declining offer {id} because there are no more tasks to launch.".format(id=offer.id.value))
+                driver.declineOffer(offer.id, self.offer_decline_filter)
+                continue
```
keymone: i'd move this out of `for offer in offers` loop and decline everything and suppress further offers until new tasks arrive

task_processing/plugins/mesos/execution_framework.py
sagar8192: Moved the state to a single dict tracking all the metadata of tasks.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Done! I am a little worried about the impact of suppressing and reviving offers on DRF algorithm of mesos. But, let's try it out.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -414,9 +204,10 @@ def create_new_docker_task(self, offer, task_config):
         command.value = task_config.cmd
 
         task.command.MergeFrom(command)
-        task.task_id.value = task_config.task_id
+        task.task_id.value = str(self.task_launch_counter)
```
sagar8192: I am ok with uuid.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +267,95 @@ def create_new_docker_task(self, offer, task_config):
...
     ####################################################################
     #                   Mesos driver hooks go here                     #
     ####################################################################
 
     def slaveLost(self, drive, slaveId):
-        log.error("Slave lost: {id}".format(id=str(slaveId)))
+        log.warning("Slave lost: {id}".format(id=str(slaveId)))
 
     def registered(self, driver, frameworkId, masterInfo):
-        self.framework_id = frameworkId.value
```
sagar8192: This value is not being used anywhere else.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Fixed!

task_processing/plugins/mesos/execution_framework.py
sagar8192: I am also unsure about how to pass driver reference to functions that need it. :(

.pre-commit-config.yaml
```diff
@@ -18,6 +18,7 @@
         files: ^tests/.+\.py$
         language_version: python3.6
     -   id: flake8
+        args: [--max-line-length=120]
```
keymone: please no :)

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +282,109 @@ def create_new_docker_task(self, offer, task_config):
...
-            paasta_print(
-                'Declining all offers since we started reconciliation '
-                'too recently'
+        for offer in offers:
+            if offer.slave_id.value in self.blacklisted_slaves:
+                log.critical("Ignoring offer {offer_id} from blacklisted slave {slave_name}".format(
+                    offer_id=offer.id.value,
+                    slave_name=offer.slave_id.value
+                )
+                )
```
Rob-Johnson: please move these brackets onto the same line

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +282,109 @@ def create_new_docker_task(self, offer, task_config):
...
+                driver.declineOffer(offer.id, self.offer_decline_filter)
+                continue
+
+            if not self.offer_matches_pool(offer):
+                log.info("Declining offer {id} because it is not for pool {pool}.".format(
+                    id=offer.id.value,
+                    pool=self.pool
+                )
+                )
```
Rob-Johnson: and again

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +13,118 @@
...
-    def on_success(self, task_id, callback):
-        pass
-
-    def on_failure(self, task_id, callback):
-        pass
+        log.info('Blacklisting slave: {id} for {secs} seconds.'.format(
+            id=slave_id,
+            secs=self.slave_blacklist_timeout_s
+        )
+        )
```
Rob-Johnson: please put these brackets on the same line

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +13,118 @@
...
+                    self.blacklisted_slaves[slave_id] +
+                    self.slave_blacklist_timeout_s
+                ):
+                    log.info('Unblacklisting slave: {id}'.format(id=slave_id))
+                    self.blacklisted_slaves.pop(slave_id, None)
+            time.sleep(10)
+
+    def enqueue_task(self, task):
+        # TODO: Add a wrapper for this task
+        self.task_metadata[task.task_id] = {
```
Rob-Johnson: can you encode this dict in a namedtuple?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +282,109 @@ def create_new_docker_task(self, offer, task_config):
...
-        # if task_params.mesos_task_state not in LIVE_TASK_STATES:
-            # task_params.marked_for_gc = True
-            # with self.constraint_state_lock:
-                # update_constraint_state(task_params.offer, self.constraints,
-                # self.constraint_state, step=-1)
+        log.info("Task update {update} received for task {task}".format(
+            update=mesos_pb2.TaskState.Name(update.state),
+            task=task_id
+        )
+        )
```
Rob-Johnson: brackets please

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +282,109 @@ def create_new_docker_task(self, offer, task_config):
...
 
+        if update.state == mesos_pb2.TASK_RUNNING:
+            self.task_metadata[task_id]['task_state'] = 'in-flight'
+
+        if update.state == mesos_pb2.TASK_FINISHED:
+            self.task_update_queue.put(self.translator(update))
```
Rob-Johnson: why're we not putting all events on the queue? the translator is the layer that decides how they're rolled up or discarded

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -481,110 +282,109 @@ def create_new_docker_task(self, offer, task_config):
...
+
+        if update.state in (
+            mesos_pb2.TASK_LOST,
+            mesos_pb2.TASK_KILLED,
+            mesos_pb2.TASK_FAILED,
+            mesos_pb2.TASK_ERROR
+        ):
+            if self.task_metadata[task_id]['retries'] >= self.task_retries:
+                log.info(
+                    'All the retries for task {task} are done.'.format(task=task_id))
```
Rob-Johnson: do we want to count against retries if the reason for the error was the staging timer failing?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +13,118 @@
...
-
 class ExecutionFramework(mesos.interface.Scheduler):
     def __init__(
         self,
-        name,
-        staging_timeout,
-        reconcile_backoff=1,
-        reconcile_start_time=float('inf'),
-        dry_run=False,
+        name="remote_run",
```
sagar8192: Done.

task_processing/plugins/mesos/execution_framework.py
sagar8192: namedtuples are immutable which poses a problem. We need to modify fields.

keymone: let's use PRecord like with events: https://github.com/Yelp/task_processing/blob/master/task_processing/interfaces/event.py

or task configs: https://github.com/Yelp/task_processing/pull/9

task_processing/plugins/mesos/execution_framework.py
sagar8192: Fixed the indentation.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Fixed the indentation.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Fixed the indentation.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Fixed the indentation.

.pre-commit-config.yaml
sagar8192: Fixed.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +13,118 @@
...
-        paasta_print(
-            'Freezing the scheduler. Further status updates and resource '
-            'offers are ignored.'
-        )
-        self.frozen = True
-        paasta_print("Killing any remaining live tasks.")
-        for task, parameters in self.tasks_with_flags.items():
-            if parameters.mesos_task_state in LIVE_TASK_STATES:
-                self.kill_task(driver, task)
+    def kill_tasks_stuck_in_staging(self):
```
sagar8192: I was thinking if we launch 1000 containers, that would mean we will spawn 1000 new threads.  Since the task state is stored in a single datastructure, the threads will have to acquire locks(going to add a mutex for all the shared datastructures). Seems like an unnecessary overhead. We can achieve the same thing with a single thread and up to 10 sec delay.

task_processing/plugins/mesos/execution_framework.py
sagar8192: I am not sure I understand the need for PRecord. We need something mutable. We can even have a class:

class TaskMetada:
        def __init__(self, task_state, retries=0, slave_name, task_config):
            self.task_state = task_state
            self.retries = retries
            slave_name = slave_name
            task_config = task_config

Then task_metadata will look like:
self.task_metadata[task_id] = TaskMetadata(...)

task_processing/plugins/mesos/execution_framework.py
sagar8192: I am not sure. In seagull, we don't distinguish between infra failure and app failure. Used the same thing here.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -15,338 +13,118 @@
...
 
-        # Gets set when registered() is called
-        self.framework_id = None
+        self.start_new_thread(self.kill_tasks_stuck_in_staging)
+        self.start_new_thread(self.unblacklist_slaves)
 
-        self.blacklisted_slaves = set()
-        self.blacklisted_slaves_lock = threading.Lock()
-        self.blacklist_timeout = 10
+    def start_new_thread(self, func):
```
sagar8192: Yeah, you are right! This does not have to be a class method. Moved it out of the class. I agree we are not saving a lot of LOC with this method, but ¯\_(ツ)_/¯

task_processing/plugins/mesos/execution_framework.py
sagar8192: We need this to honor implicit retries. The translator does not know about retries. So, it cannot make a decision about whether to forward the event or not. Right now we only forward events if we receive mesos_pb2.TASK_FINISHED or the number of retries are over.

task_processing/plugins/mesos/execution_framework.py
jolynch: +1 to using a type checking datastructure here, but I'm also cool with punting to next iteration.

sagar8192: Discussed with jlynch offline. Created this https://jira.yelpcorp.com/browse/TASKPROC-28 to fix it.

jolynch:
I'm going to merge this and we can keep iterating.

task_processing/plugins/mesos/execution_framework.py
keymone: something like retry_policy would be a task_executor-level option, not execution framework imo. this way if somebody doesn't like our executor they can implement another one and be able to deal with all events without having to touch mesos-related code. lets just publish all events into that queue and have executor/runner deal with them (and deal with retries).

execution framework would only be a place for dealing with mesos offers, reconciliations, failover, etc.

PR #10, opened 2017-05-18
get the async example working

Rob-Johnson:
also make the event handler a namedtuple


task_processing/runners/async.py
```diff
@@ -1,9 +1,11 @@
+from collections import namedtuple
 from threading import Thread
 
 from six.moves.queue import Empty
 
 from task_processing.interfaces.runner import Runner
 
+EventHandler = namedtuple('EventHandler', ['matcher_func', 'cb'])
```
keymone: maybe `predicate` instead of `matcher_func`?

task_processing/plugins/mesos/translator.py
```diff
@@ -2,17 +2,17 @@
 
 # https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto
 MESOS_TASK_STATUS_TO_EVENT = {
-    0: Event(platform_type='starting'),
-    1: Event(platform_type='running'),
-    2: Event(platform_type='finished'),
-    3: Event(platform_type='failed'),
+    0: Event(platform_type='starting', terminal=False),
```
keymone: make `initial=False` in field definition

examples/async.py
```diff
@@ -16,28 +16,29 @@ def __init__(self):
...
     mesos_address = os.environ['MESOS']
     executor = MesosExecutor(
-        credential_secret_file="/src/task_processing/examples/cluster/secret",
+        credential_secret_file="/src/examples/cluster/secret",
         mesos_address=mesos_address
     )
 
     counter = Counter()
-    runner = Async(executor, [(EventTerminal, counter.process_event)])
+    runner = Async(executor, [EventHandler(matcher_func=lambda x: x.terminal, cb=counter.process_event)])
```
keymone: 80 chars

task_processing/plugins/mesos/translator.py
Rob-Johnson: why? I don't think there is a reasonable default - I'd leave it to the implementer to decide either way per Event?

keymone: should `None` mean anything different from `False` then? if it does, we need to make sure an error is thrown when such event ends up in `if event.terminal` (make an object with __bool__ method that throws an exception). if it doesn't - then let's at least make the field mandatory?

Rob-Johnson: | should None mean anything different from False then?

You shouldn't be able to do ``None``? maybe I can provide a predicate like: ```lambda x: x in ['True', 'False']``` if the typing isn't strong enough.

|  if it doesn't - then let's at least make the field mandatory?
agreed

keymone:
ship.

PR #11, opened 2017-05-18
Registry class that allows us to dynamically alter TaskProc behavior

jolynch:
With this we can dynamically alter the capabilities of the libraries
using customer provided plugins!


jolynch:
I don't like walking everything, but walking a specified package seems fine to me (similar to pyramid's .[scan](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.scan) method). I'd probably want to do that as a next iteration though.

I am not sure I agree about the explicitness being necessarily bad. With this the PaaSTA diff to support, e.g., AWS batch would be adding the task_processing_aws_batch package to the setup.py and doing:

```python
task_processing_registry.load_plugin('task_processing_aws_batch')
```
This is better than manually importing because otherwise you'd have to make PaaSTA code changes to import the class and do their own bookeeping.

task_processing/registry.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # Later...
+        executor = reg.executor_from_config(
+            provider='foobar',
+            provider_config={'specific': 'config', 'for', 'plugin'}
+        )
+        executor.run()
+    """
+
+    plugin_registry = pyrsistent.m()
```
jolynch: Yea I just like the pyrsistent datastructures, makes me think about mutating them :-P

/me shrug I don't have strong opinions.

task_processing/registry.py
```diff
@@ -0,0 +1,93 @@
...
+        reg.load_plugin('my.plugin')
+        # ... within the my.plugin.register_plugin method:
+        registry.add_task_executor('foobar', FooBarExecutor)
+
+        # register 'another'
+        reg.load_plugin('another.plugin')
+
+        # Later...
+        executor = reg.executor_from_config(
+            provider='foobar',
```
jolynch: I do want plugins to be able to supply multiple executors, but prefixing them with the plugin name (and making that the package name) would have the nice side effect of making sure you can't have two of the same name ... that's actually pretty nice.

I will change this tomorrow to prefix all registered executors with the package name (which is the plugin name)?

task_processing/registry.py
keymone: ack. if somebody really hates it - we can discuss further. shipit.

task_processing/registry.py
keymone: hmm.. i'm not sure about plugin providing multiple executors. package providing multiple plugins - sure.

are you thinking about this:

    plugins/
      mesos/
        mesos_api_executor.py
        mesos_http_executor.py

or this:

    plugins/
      mesos_api/
        executor.py
      mesos_http/
        executor.py

?

keymone:
right, so in paasta service configs we'd have something like

    ---
    task_processing_provider: task_processing.plugins.mesos

and in paasta_taskproc_run.py (or whatever) we won't need to explicitly `import task_processing.plugins.mesos` but instead will run `registry.load_plugin(svc_cfg['task_processing_provider'])`?

sounds good to me. shipit.

PR #14, opened 2017-05-19
task metadata prec

keymone:
wrap task metadata into type-checked TaskMetadata precord

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -5,6 +5,8 @@ import time
 import mesos.interface
 import mesos.native
 from mesos.interface import mesos_pb2
+from pyrsistent import field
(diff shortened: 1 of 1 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -16,6 +18,14 @@
 log = logging.getLogger(__name__)
 
 
+class TaskMetadata(PRecord):
+    slave_id = field(type=str, initial=None)
+    retries = field(type=int, initial=0)
+    task_config = field(type=PRecord, mandatory=True)
+    task_state = field(type=str, initial='enqueued')
```
Rob-Johnson: In the long term, I'd like to make ```State``` a real thing, similar to Event.

keymone: agree

PR #15, opened 2017-05-19
extract retrying behavior out into separate executor

keymone:
also add `success` and `task_config` attributes to each event

### DIFF
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -3,8 +3,6 @@ from pyrsistent import m
 from pyrsistent import PMap
 from pyrsistent import PRecord
 
-from task_processing.interfaces.task_executor import TaskConfig
-
 
 class Event(PRecord):
     # reference to platform-specific event object
@@ -16,5 +14,5 @@ class Event(PRecord):
     platform_type = field(type=str)
     # task_id this event pertains to
     task_id = field(type=str)
-    task_config = field(type=TaskConfig)
-    middleware_data = field(type=PMap, initial=m())
+    task_config = field(type=PRecord)
+    extensions = field(type=PMap, initial=m())
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -1,7 +1,9 @@
 from pyrsistent import field
+from pyrsistent import m
+from pyrsistent import PMap
 from pyrsistent import PRecord
 
-# TODO: organize and explain these
+from task_processing.interfaces.task_executor import TaskConfig
 
 
 class Event(PRecord):
@@ -9,7 +11,10 @@ class Event(PRecord):
     raw = field()
     # is this the last event for a task?
     terminal = field(type=bool, mandatory=True)
+    success = field(type=(bool, type(None)), initial=None)
     # platform-specific event name
     platform_type = field(type=str)
     # task_id this event pertains to
     task_id = field(type=str)
+    task_config = field(type=TaskConfig)
+    middleware_data = field(type=PMap, initial=m())
--- a/task_processing/interfaces/task_executor.py
+++ b/task_processing/interfaces/task_executor.py
@@ -51,7 +51,7 @@ class TaskExecutor(object):
         pass
 
     @abc.abstractmethod
-    def event_queue(self):
+    def get_event_queue(self):
         """Get queue of events
 
         :param str task_id: The task that you want to get status on
--- a/task_processing/interfaces/task_executor.py
+++ b/task_processing/interfaces/task_executor.py
@@ -45,8 +45,14 @@ class TaskExecutor(object):
         pass
 
     @abc.abstractmethod
-    def status(self, task_id):
-        """Get Status of the specified task
+    def stop(self):
+        """Stop the executor stack
+        """
+        pass
+
+    @abc.abstractmethod
+    def event_queue(self):
+        """Get queue of events
 
         :param str task_id: The task that you want to get status on
 
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -8,6 +8,8 @@ from pyrsistent import field
 from pyrsistent import PRecord
 from six.moves.queue import Queue
 
+from task_processing.plugins.mesos.translator import mesos_status_to_event
+
 
 FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s - %(message)s'
 LEVEL = logging.DEBUG
@@ -28,10 +30,10 @@ class ExecutionFramework(Scheduler):
         self,
         name,
         role,
-        translator,
         task_staging_timeout_s=240,
         pool=None,
         max_task_queue_size=1000,
+        translator=mesos_status_to_event,
         slave_blacklist_timeout_s=900,
         offer_backoff=240,
         task_retries=2,
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -8,8 +8,6 @@ from pyrsistent import field
 from pyrsistent import PRecord
 from six.moves.queue import Queue
 
-from task_processing.plugins.mesos.translator import mesos_status_to_event
-
 
 FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s - %(message)s'
 LEVEL = logging.DEBUG
@@ -30,10 +28,10 @@ class ExecutionFramework(Scheduler):
         self,
         name,
         role,
+        translator,
         task_staging_timeout_s=240,
         pool=None,
         max_task_queue_size=1000,
-        translator=mesos_status_to_event,
         slave_blacklist_timeout_s=900,
         offer_backoff=240,
         task_retries=2,
@@ -361,7 +359,10 @@ class ExecutionFramework(Scheduler):
             task=task_id
         ))
 
-        self.task_update_queue.put(self.translator(update, task_id))
+        md = self.task_metadata[task_id]
+        self.task_update_queue.put(
+            self.translator(update, task_id).set(task_config=md.task_config)
+        )
 
         if update.state == 'TASK_FINISHED':
             self.task_metadata.pop(task_id, None)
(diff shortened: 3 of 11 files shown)


task_processing/interfaces/event.py
```diff
@@ -1,13 +1,18 @@
...
 
 class Event(PRecord):
     # reference to platform-specific event object
     raw = field()
     # is this the last event for a task?
     terminal = field(type=bool, mandatory=True)
+    success = field(type=(bool, type(None)), initial=None)
     # platform-specific event name
     platform_type = field(type=str)
+    task_config = field(type=TaskConfig)
```
jolynch: I'm pretty confused what we're trying to store in the Event records.

What are task_config? What are middleware_data, what do we mean by platform?

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -0,0 +1,60 @@
+import time
+from threading import Thread
+
+from pyrsistent import m
+from six.moves.queue import Queue
+
+from task_processing.interfaces.task_executor import TaskExecutor
+
+
+class RetryingExecutor(TaskExecutor):
```
jolynch: Why do we need retrying at all? Can the user just hit run(<config>) again if they get back "failed" from status?

task_processing/interfaces/event.py
keymone: yeah, that's not ideal. task config at least isn't. it's sort of a tradeoff between having events passing more related stuff around or having to track a lot of data per executor "layer".

which bring me to executor "layers" - this (RetryingExecutor) is an attempt to take an aspect of execution framework that is orthogonal to actually running tasks and extract it into something that is composable. i've settled on executors to be composable, because they all implement the same interface: run/kill/stop, maybe it's wrong place to do it, but having composable layers is useful imo.

middleware_data is a place where different "middleware" (composable execution layers) can store some data useful for end-user. possibly bad naming. naming is hard.

task_processing/plugins/mesos/retrying_executor.py
keymone: it's a matter of conveniece? the argument "why provide `function` if user can just do `implementation body`" always rests on a tradeoff between flexibility and convenience.

task_processing/interfaces/event.py
Rob-Johnson: I think it's overkill to store the whole TaskConfig object - can the event just have a task_id instead?

task_processing/interfaces/event.py
```diff
@@ -1,13 +1,18 @@
...
+    middleware_data = field(type=PMap, initial=m())
```
Rob-Johnson: im also not really convinced by this. I kind of get that you need to store data somewhere, but the event seems the wrong place to do it.

task_processing/plugins/mesos/retrying_executor.py
Rob-Johnson: I think the same argument could be made for any of the executors. I guess we're just providing a set of implementations for common patterns for running tasks. I think this is fine.

Rob-Johnson: I'd kind of expect this behavior to be at the ``Runner`` layer, rather than ``Executor``? (I'd expect this to have looked very similar to the ``AsyncRunner``, with the ``predicate`` matching on non-success && terminal events, and the ``cb`` ``executor.run``)?

keymone: i did consider runners, problem is they aren't composable. executors are due to very simple and strict interface.

task_processing/interfaces/event.py
keymone: the reason we need task_config is because retrying logic has to call .run with task_config so id is not enough. we could store id->config map in retrying executor which can be wasteful.

personally i don't have a strong feeling that task_config shouldn't be part of the event structure. there is only so much data that is related to the task, and it's source configuration that it was launched with - one of the most important things about it. it allows for all sorts of heuristics to be implemented in the composable layers without each having to store a crapload of data.

task_processing/interfaces/event.py
keymone: i don't think there is a better place for it actually. events are emitted from lowest level executor up to the upper layers and they get enriched with useful information along the way. we could say "why not just emit the pair of (event, mutable state) and let upper layers mutate the state rather than the event" but i would argue that it's already pretty much what we're doing with Event precord - only the `raw` is actual event, the rest is executor-provided data that helps upper layers (or end-user) to work with it.

task_processing/plugins/mesos/retrying_executor.py
sagar8192: I am also not sure whether we need this. I think we can have either have implicit retries(in the scheduler) or let clients decide if they want to retry.

keymone: that leads to god-classes with way too many responsibilities

keymone:
@jolynch @sagar8192 @Rob-Johnson 

let's make a decision here. i think we already have consensus that executors are the composable building blocks so if there are any cosmetic changes required to get this merged - tell me. if you think this is wrong - still tell me :)

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -361,7 +361,10 @@ def statusUpdate(self, driver, update):
             task=task_id
         ))
 
-        self.task_update_queue.put(self.translator(update, task_id))
+        md = self.task_metadata[task_id]
```
jolynch: This is a really nasty abstraction leak. I think we need a better way to communicate events and status updates up and down the executor stack. I like the idea of a queue of Events, but it probably needs to be multi-consumer single producer or some such. In java they have Event Listeners and notifiers, maybe we want something similar?

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -0,0 +1,60 @@
...
+                    ('middleware_data', 'retries'),
+                    self.retries - self.task_retries[e.task_id])
+                self.dest_queue.put(et)
+
+            if self.stopping:
+                return
+
+            time.sleep(1)
+
+    def run(self, task_config):
```
jolynch: This executors TASK_CONFIG_INTERFACE needs to be the wrapped TASK_CONFIG_INTERFACE

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -0,0 +1,60 @@
...
+            while not self.src_queue.empty():
+                e = self.src_queue.get()
+                if e.terminal and self.retry_pred(e):
+                    if self.task_retries[e.task_id] > 0:
+                        self.run(e.task_config)
+                        self.task_retries = self.task_retries.set(
+                            e.task_id, self.task_retries[e.task_id] - 1)
+                        continue
+                et = e.transform(
+                    ('middleware_data', 'retries'),
```
jolynch: maybe extensions and key by the extension?

e.g.
```
{'extensions':
    'retrying': ...
}
```
I'm not a big fan of "middleware_data", I don't really know what it means.

task_processing/interfaces/event.py
```diff
@@ -1,15 +1,20 @@
 from pyrsistent import field
+from pyrsistent import m
+from pyrsistent import PMap
 from pyrsistent import PRecord
 
-# TODO: organize and explain these
+from task_processing.interfaces.task_executor import TaskConfig
 
 
 class Event(PRecord):
```
jolynch: I think we're going to start making this bigger and bigger, what are the things that really need to be here?

task_processing/interfaces/event.py
```diff
@@ -1,15 +1,20 @@
 from pyrsistent import field
+from pyrsistent import m
+from pyrsistent import PMap
 from pyrsistent import PRecord
 
-# TODO: organize and explain these
+from task_processing.interfaces.task_executor import TaskConfig
```
jolynch: I don't think we want this, just a PRecord.

task_processing/interfaces/event.py
keymone: `middleware_data` is an attempt to constrain the growth. we definitely need this record to be extensible without having to modify the source.

task_processing/plugins/mesos/execution_framework.py
keymone: i wonder why do you think it's an abstraction leak: we're not exposing some internal detail here, we're simply attaching the actual input the user had passed to us. it's like a `source` field, helps understanding what caused the event.

task_processing/plugins/mesos/retrying_executor.py
keymone: i guess i borrowed the name from my rails experience where apps are built as a stack of middleware classes. anyway, `extension` seems more sensible, i'll do that.

task_processing/plugins/mesos/retrying_executor.py
keymone: i don;t understand, could you give an example?

jolynch: In this TaskExecutor's __init__ method:

```
self.TASK_CONFIG_INTERFACE = executor.TASK_CONFIG_INTERFACE
```

TASK_CONFIG_INTERFACE is the way that callers of TaskExecutors know what parameters are needed by run.

jolynch:
After chatting with @Rob-Johnson and @keymone I think the following is the consensus (correct me if I'm wrong):

We're doing the following:

1. All TaskExecutors _must_ implement:

```python
@abstractproperty
def event_queue() -> Queue
```

2. All of these queues _should_ be fixed size so as to prevent overflow
3. Runners _must_ always consume events from their downstream executors
4. If we need complex topologies, we will create a TaskExecutor (such as FanoutExecutor) which dynamically creates queues and publishes downstream events onto those queues as needed.

task_processing/interfaces/task_executor.py
```diff
@@ -45,8 +45,14 @@ def kill(self, task_id):
...
     @abc.abstractmethod
-    def status(self, task_id):
-        """Get Status of the specified task
+    def stop(self):
+        """Stop the executor stack
+        """
+        pass
+
+    @abc.abstractmethod
+    def event_queue(self):
```
Rob-Johnson: could you make this ``get_event_queue``?

keymone: i don't really like that naming scheme but ok

PR #16, opened 2017-05-19
Initial Doc Support (TASKPROC-29)

nhandler:
`make tox_docs` or `make docs` both produce Sphinx documentation that is generated from our Python scripts. I have intentionally included the generated documents (only the `rst` versions) in this commit to match the style of other Yelp projects.


Makefile
```diff
@@ -13,6 +13,9 @@ dev_env:
 tox_%: dev_env
 	${TOX} -e $*
 
+docs: dev_env
```
jolynch: do we need dev_env for docs?

nhandler: Without it, you don't have `tox`.

PR #17, opened 2017-05-23
Use pymesos

jolynch:
This branch changes us from using mesos.native to pymesos, which unblocks python3 compatibility and allows us to actually run our integration tests.

@sagar8192 ++ for the help with this.


task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -103,5 +101,8 @@ def stop(self):
         self.driver.stop()
         self.driver.join()
 
+    def status(self):
```
sagar8192: Do we need this? Currently, all the events are only shared via task_update_queue. We don't store any state in the scheduler, so it will be difficult to implement it.

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -64,28 +63,27 @@ def __init__(self,
...
-            else:
-                with open(credential_secret_file) as f:
-                    credential.secret = f.read().strip()
+            with open(credential_secret_file) as f:
+                secret = f.read().strip()
 
         self.execution_framework = ExecutionFramework(
             name="test",
-            staging_timeout=10,
+            task_staging_timeout_s=30,
```
sagar8192: How about 60 secs instead?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -306,6 +317,8 @@ def slaveLost(self, drive, slaveId):
         log.warning("Slave lost: {id}".format(id=str(slaveId)))
 
     def registered(self, driver, frameworkId, masterInfo):
+        if self.driver is None:
```
sagar8192: ++ This is a better place to initialize the driver.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -133,28 +131,27 @@ def enqueue_task(self, task_config):
             log.info('Reviving offers because we have tasks to run.')
 
     def build_framework_info(self):
-        framework = mesos_pb2.FrameworkInfo()
-        framework.user = ""  # Have Mesos fill in the current user.
+        framework = Dict()
+        # TODO: why does this need to be root, can it be "mesos plz figure out"
+        framework.user = 'root'
```
Rob-Johnson: if this isn't root, then the slave wont run tasks. you need the framework to be root for the slave to run tasks as root.

task_processing/plugins/mesos/mesos_executor.py
Rob-Johnson: +1 - it can take quite a while to pull big (yelp-main) docker images.

task_processing/plugins/mesos/mesos_executor.py
jolynch: This is just implementing an interface that we previously defined. We can discuss more on TASKPROC-30

task_processing/plugins/mesos/execution_framework.py
jolynch: Yea, this had to be root for the tests to pass, but I'd love for mesos to just launch tasks as whatever user launched them. Sagar mentioned that another option is having the docker socket owned by users, but I think kwa had issues with that.

PR #18, opened 2017-05-23
task building cleanup

keymone:
the code for building a task bothered me so i've cleaned it up and restructured to be less imperative. also the build_* functions are gone because they're now just single Dict() calls.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -40,19 +40,21 @@ class ExecutionFramework(Scheduler):
         self.name = name
         # wait this long for a task to launch.
         self.task_staging_timeout_s = task_staging_timeout_s
-        self.framework_info = self.build_framework_info()
         self.pool = pool
         self.translator = translator
         self.slave_blacklist_timeout_s = slave_blacklist_timeout_s
         self.offer_backoff = offer_backoff
         self.task_retries = task_retries
 
+        # TODO: why does this need to be root, can it be "mesos plz figure out"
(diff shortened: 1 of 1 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -40,19 +40,21 @@ def __init__(
...
-        self.framework_info = self.build_framework_info()
         self.pool = pool
         self.translator = translator
         self.slave_blacklist_timeout_s = slave_blacklist_timeout_s
         self.offer_backoff = offer_backoff
         self.task_retries = task_retries
 
+        # TODO: why does this need to be root, can it be "mesos plz figure out"
+        self.framework_info = Dict(
+            user='root', name=self.name, checkpoint=True)
```
sagar8192: nit: Can we have each argument on a separate line? I find it easier to read. But, nbd.

self.framework_info = Dict(
      user='root', 
      name=self.name, 
      checkpoint=True
)

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -217,94 +208,47 @@ def get_tasks_to_launch(self, offer):
...
-        # Set docker container in task.container
-        task.container = container
 
-        return task
+        return Dict(
+            task_id=Dict(value=task_config.task_id),
+            agent_id=Dict(value=offer.agent_id.value),
+            name='executor-{id}'.format(id=task_config.task_id),
+            resources=[
+                Dict(name='cpus',
```
sagar8192: Same comment as above about indentation.

PR #21, opened 2017-05-25
Add support for mesos role in execution framework

sagar8192:
How does this work?
All the frameworks must use a role while registering with mesos. The framework will only use resources that belong to their role. Mesos will offer them resources with role * and their own role.

How do I know it works?
Ran all the examples and `make test`

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -29,6 +29,7 @@ class ExecutionFramework(Scheduler):
     def __init__(
         self,
         name,
+        role,
         task_staging_timeout_s=240,
         pool=None,
         max_task_queue_size=1000,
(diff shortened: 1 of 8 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -223,15 +231,19 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
                      type='SCALAR',
+                     role=self.role,
                      scalar=Dict(value=task_config.mem)),
                 Dict(name='disk',
                      type='SCALAR',
+                     role=self.role,
                      scalar=Dict(value=task_config.disk)),
                 Dict(name='ports',
                      type='RANGES',
+                     role=self.role,
```
Rob-Johnson: im surprised we have to add the role to every resource

PR #22, opened 2017-05-29
[wip] mostly naïve test coverage

keymone:


requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Library so -e .
 -e .
+pymesos
```
jolynch: we probably don't want pymesos here, instead in a separate requirements-mesos_plugin.txt:

```
-e .[mesos_executor]
```
Or something. Tbh I'm sorta fine with a separate tox target (see the mesos tox target) that has that dependency there.

tox.ini
```diff
@@ -10,9 +10,9 @@ deps =
     -rrequirements.txt
     -rrequirements-dev.txt
 commands =
-    pytest -v {posargs:tests}/unit
     pre-commit install -f --install-hooks
     pre-commit run --all-files
+    pytest --cov=task_processing --cov-fail-under=80 -v {posargs:tests}/unit
```
jolynch: ambitious, maybe set to 30 and we can bump it iteratively?

PR #23, opened 2017-06-14
Add function to make data structures thread safe

charleskwwan:
# Review Request (ref. TASKPROC-27)
## Description
* Add utility function make_thread_safe_struct.
* Add test to ensure thread safe context is active during method calls.
### Manual
* Ran provided test in examples with a Mesos cluster; tasks complete as expected.
* Ran function on own threads to test thread safety.
### Unit
* Created unit test to test that lock is initialized, and context is active during relevant calls, but not during irrelevant calls.
## Note to reviewers
* Unit tests do not generate their own threads, as it was assumed that this would be too difficult to simulate realistically.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -61,7 +61,7 @@ class ExecutionFramework(Scheduler):
         self.are_offers_suppressed = False
 
         self.offer_decline_filter = Dict(refuse_seconds=self.offer_backoff)
-        # TODO: These should be thread safe
+        self._lock = threading.RLock()
         self.blacklisted_slaves = {}
         self.task_metadata = {}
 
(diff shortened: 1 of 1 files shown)


keymone:
why do we not simply use pysistent map/precord?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -55,17 +57,28 @@ def __init__(
             checkpoint=True,
             role=self.role
         )
-        self.task_queue = Queue(max_task_queue_size)
-        self.task_update_queue = Queue(max_task_queue_size)
+
+        self.task_queue = make_thread_safe_struct(
```
jolynch: Hm, I'm concerned about this. Afaik Python [Queues](https://docs.python.org/3/library/queue.html#module-queue) _are_ thread safe (as are [collections.dequeue](https://docs.python.org/3/library/collections.html#collections.deque)). We don't need to add locking on top of them IMO unless we have a very specific re-entrant case. If we need multi-process support multiprocessing has a Queue equivalent which is quite good too.

I'm a much bigger fan of using pyrsistent datastructures everywhere and for the pubsub using well established stdlib queues (or a package). I don't think we want to be implementing our own locked datastructures.

charleskwwan: Pyrsistent seems like a good option, and I agree there isn't really a need to add locking on top of queues. 

However, I am concerned about how non-queue data structures like dictionaries/map would prevent race conditions. I suppose they are 'thread-safe' in the sense that since threads gain copies, they do not modify the original and thus do not pollute another thread's work. However, should they try to commit changes on the same thing, you get a conflict. As far as I know, pyrsistent doesn't solve this synchronicity issue itself. I think I'd need to implement locking here.

Alternatively, the thought had crossed my mind that if we instead convert to a coroutine model, explicit async programming would make spotting and controlling race conditions much easier. This way, we could avoid worrying about the nature of the data structures themselves, only about when and how they are used.

Thoughts?

jolynch: Immutable data structures don't help with mutable races (e.g. the queues, that's what locking is for), but they do help limit the mutability within those mutables stores which limits races.

For example, imagine that you have a queue of dicts, and someone peeks at the top dict and modifies the 'done' key. In the pyrsistent world that can't happen, so you don't have to worry about other threads of execution messing with your state. You must instead consume the item from the queue and add a new item back on, which is as previously mentioned thread safe (because of the Queue internal locking).

We definitely want to move to a coroutine model IMO, which is part of the reason we're moving all the things to Python3 (and the faster we use gevent or asyncio the better IMO), but honestly you end up with the same problems since Python threads can't actually run concurrently (same as green threads or coroutines); you still have to lock critical sections when you have coroutines.

PR #26, opened 2017-06-20
Kawaiwan make thread safe

charleskwwan:
# Description
- Added locks to execution framework
- Converted blacklisted_slaves and task_metadata from python dictionaries to pyrsistent pmaps
# Testing done
- Manual testing to check locks acquire and release
- Manual testing to check that persistent data structures are being re-set appropriately after changes
- Ran itest in examples
# Note to reviewers
- As per discussion with jolynch and sagarp, previous attempt using a function to metaclass data structures to enforce thread safety would not work as intended. In order to control race conditions, I made the above changes.
- Future developers on this code will need to be wary of the locked contexts, unfortunately.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -35,8 +35,8 @@ class ExecutionFramework(Scheduler):
         max_task_queue_size=1000,
         translator=mesos_status_to_event,
         slave_blacklist_timeout_s=900,
-        offer_backoff=240,
-        task_retries=2,
+        offer_backoff=10,
+        suppress_delay=10,
     ):
         self.name = name
         # wait this long for a task to launch.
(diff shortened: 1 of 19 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -102,37 +106,50 @@ def kill_task(self, task_id):
...
+        with self._lock:
+            self.task_metadata = self.task_metadata.set(
+                task_config.task_id,
+                TaskMetadata(
+                    task_config=task_config,
+                    time_launched=time.time(),
+                )
+            )
+
         self.task_queue.put(task_config)
```
sagar8192: This should also be locked?

charleskwwan: Agreed.

PR #27, opened 2017-06-21
tweaks in credentials passing and mesos exec framework initialization

keymone:
- shift credentials concern
- allow overriding framework kwargs in mesos executor

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -56,8 +56,9 @@ class MesosExecutor(TaskExecutor):
         principal='taskproc',
         secret=None,
         mesos_address='127.0.0.1:5050',
-        translator=mesos_status_to_event,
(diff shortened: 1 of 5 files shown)


task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -53,10 +53,11 @@ class MesosExecutor(TaskExecutor):
     def __init__(
         self,
         role,
-        authentication_principal='taskproc',
-        credential_secret_file=None,
+        principal='taskproc',
+        secret=None,
         mesos_address='127.0.0.1:5050',
         translator=mesos_status_to_event,
+        framework_kwargs={},
```
Rob-Johnson: in general I'm in favour of being explicit with parameters, rather than asking the caller to go looking through the code to find out what parameters they can set. I'd make each of the keys you set the default for explicit in here.

PR #29, opened 2017-06-22
Fixes to exec framework, retrying exec, sync runner

keymone:
- build docker parameters in exec framework
- cleanup leftovers of retry behavior in exec framework
- lock task retries map in retrying exec
- add logging to retrying exec
- cleanup task retries map after task is finished
- add success field to TASK_FINISHED event translator
- drop verbosity of sync runner

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -411,7 +411,7 @@ class ExecutionFramework(Scheduler):
             'TASK_FINISHED'
         ):
             with self._lock:
-                self.task_metadata.pop(task_id, None)
+                self.task_metadata = self.task_metadata.discard(task_id)
 
         # We have to do this because we are not using implicit
         # acknowledgements.
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -411,7 +411,7 @@ class ExecutionFramework(Scheduler):
             'TASK_FINISHED'
         ):
             with self._lock:
-                self.task_metadata = self.task_metadata.discard(task_id)
+                self.task_metadata.pop(task_id, None)
 
         # We have to do this because we are not using implicit
         # acknowledgements.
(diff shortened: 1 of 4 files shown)


task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -1,11 +1,19 @@
...
+from operator import sub
+from threading import Lock
 from threading import Thread
 
 from pyrsistent import m
 from six.moves.queue import Queue
 
 from task_processing.interfaces.task_executor import TaskExecutor
 
+FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s - %(message)s'
```
Rob-Johnson: I don't think we should be doing this from inside the module.

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -15,41 +23,79 @@ def __init__(self,
...
+        self.task_retries_lock = Lock()
+
         self.src_queue = executor.get_event_queue()
         self.dest_queue = Queue()
-        self.retry_thread = Thread(target=self.retry_loop)
-        self.retry_thread.start()
         self.stopping = False
         self.TASK_CONFIG_INTERFACE = executor.TASK_CONFIG_INTERFACE
 
+        self.retry_thread = Thread(target=self.retry_loop)
```
Rob-Johnson: do you need to make this a daemon thread?

keymone: done

task_processing/plugins/mesos/retrying_executor.py
keymone: done

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -374,15 +378,6 @@ def statusUpdate(self, driver, update):
         if update.state == 'TASK_FINISHED':
             self.task_metadata.pop(task_id, None)
 
-        if update.state in (
```
jolynch: This conflicts with @huadongliu 's change I think, why are you removing it?

keymone: my mistake, i only needed to remove the `task_update_queue.put` line because we already enqueue that update in line 374. good catch!

PR #31, opened 2017-06-27
Dynamodb persistence

Rob-Johnson:


task_processing/plugins/persistence/dynamodb_persistence.py
```diff
@@ -0,0 +1,75 @@
...
+            Item=self._event_to_item(event)
+        )
+
+    def _event_to_item(self, event):
+        thawed = thaw(event)
+        resp = {}
+        for k, v in thawed.items():
+            if type(v) is str:
+                resp[k] = {
+                    'S': v
```
Rob-Johnson: you have to do this weird type encoding for dynamodb - it's part of the API. There is another interface in which boto can do this for you, but it doesn't let you add in custom serializers (like I have for datetime in the json serializing/deserializing), so I have to do this manually. You're right that we could get round this by storing timestamp as an int, but I'm not sure I really like that (other times I've done this the code has been littered with encoding int -> datetime -> int).

task_processing/plugins/persistence/dynamodb_persistence.py
```diff
@@ -0,0 +1,75 @@
...
+        self.ddb_client = session.client(
+            service_name='dynamodb',
+            endpoint_url=endpoint_url,
+        )
+        self.table = session.resource(
+            endpoint_url=endpoint_url,
+            service_name='dynamodb'
+        ).Table(table_name)
+
+    def read(self, task_id, comparison_operator='EQ'):
```
Rob-Johnson: I can imagine that if paasta encodes the task_id as service.instance.uuid then we'll do ``BEGINS_WITH`` to find all the tasks for a given service.instance

task_processing/plugins/persistence/dynamodb_persistence.py
```diff
@@ -0,0 +1,75 @@
...
+                }
+            elif type(v) is int:
+                resp[k] = {
+                    'N': str(v)
+                }
+            elif type(v) is dict:
+                resp[k] = {
+                    'M': self._event_to_item(v)
+                }
+            elif type(v) is list:
```
sjaensch: One thing I'd note is that using `type(x) is y` is kind of an antipattern in Python - I'd prefer `isinstance`, although I have no context as to where the data is coming from, so it might not make a difference.

In this specific case, you actually support any iterable, not just a list. I'd probably use something like `if isinstance(e, collections.Iterable):` here.

Rob-Johnson: ack, thanks for the tip. I'll change to use isinstance.

task_processing/plugins/persistence/dynamodb_persistence.py
keymone: yeah, ignore my comment then. as for int <-> datetime, ideally it would only need to happen once in event_to_item and item_to_event but that doesn't matter much if we can't avoid dynamodb encoding scheme part.

PR #32, opened 2017-06-29
fix task state str typecheck

keymone:
pymesos update.state is unicode, we need str

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -483,19 +483,20 @@ class ExecutionFramework(Scheduler):
         )
 
         # Record state changes
-        if md.task_state != update.state:
+        task_state = update.state.encode('utf-8')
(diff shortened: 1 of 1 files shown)


sjaensch:
Hint: this is a Python 3 codebase, so technically it's `str` and you need `bytes`. 😅

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -483,19 +483,20 @@ def statusUpdate(self, driver, update):
         )
 
         # Record state changes
-        if md.task_state != update.state:
+        task_state = update.state.encode('utf-8')
```
jolynch: Why do we want bytes here? If there is a validator just change it to str and it should work ok? (or call str on update.state)

keymone: went with calling str on update.state. ship?

PR #33, opened 2017-06-29
store timestamp as a float; remove datetime bits

Rob-Johnson:


jolynch:
Test failure? Looks related

task_processing/interfaces/event.py
```diff
@@ -23,17 +26,19 @@ class Event(PRecord):
 
 
 def json_serializer(o):
-    if isinstance(o, datetime.datetime):
-        return o.strftime('%s')
+    if isinstance(o, uuid.UUID):
+        print("encoded value of o %s is %s" % (o, o.hex))
```
keymone: drop the print?

task_processing/interfaces/event.py
```diff
@@ -23,17 +26,19 @@ class Event(PRecord):
...
+        return o.hex
+    return json.JSONEncoder.default(o)
 
 
 def json_deserializer(dct):
     for k, v in dct.items():
-        if k == 'timestamp':
+        if k == "uuid":
```
keymone: i don't see a reason to iterate over all keys if we know which specific keys need special treatment

task_processing/plugins/persistence/dynamodb_persistence.py
```diff
@@ -56,26 +59,21 @@ def _event_to_item(self, e):
...
         elif type(raw) is int:
             return {
-                'I': str(raw)
+                'N': str(raw)
             }
         else:
             print("Missed converting key %s type %s" % (raw, type(raw)))
 
+    def item_to_event(self, obj):
+        return self._replace_decimals(obj)
```
keymone: i'm guessing here we need this because type check fails on precord for int/float fields? it's probably "better" to use precord field factories instead of conflating ints and floats here.

    field(factory=xxx) <- will call xxx() on whatever value supplied to intializer

task_processing/interfaces/event.py
Rob-Johnson: because we still do a 'freeze' on the other vals to turn them into pyrsistent types

PR #34, opened 2017-06-29
Expose ExecutionFramework pool setting via MesosExecutor

huadongliu:
This is TASKPROC-52. examples/sync.py can run against a real cluster with "python examples/sync.py -s 8385iBfX5s2BrSh6eiG -r taskproc -p taskproc -m 10.40.5.5:5050". There are no confusing zero resource offers in the log any more https://pb.yelpcorp.com/286124 vs. https://pb.yelpcorp.com/286126.

### DIFF
--- a/examples/sync.py
+++ b/examples/sync.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import argparse
 import logging
 import os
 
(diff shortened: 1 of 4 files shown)


keymone:
why do we need to default pool to some value? for the sake of simplicity i'd leave it None and have users provide it if they need it. right now we'd have to update example clusters in this repo and paasta repo and document that pool is not empty by default, otherwise nothing will work.

huadongliu:
The change was meant to correct confusing mesos offer messages. It makes sense for users to provide a pool setting when running task_processing. I will make a cut.

PR #35, opened 2017-07-06
control events

keymone:
this is naive implementation of control events. they share the same datastructure as task events and some fields. we need them because we need a way to do controlled crash-landing in case of unrecoverable incidents (like framework being remvoed from mesos).

LF discussion on implementation details

https://github.com/Yelp/paasta/pull/1325 is blocked on this

### DIFF
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -8,9 +8,13 @@ from pyrsistent import PMap
 from pyrsistent import PRecord
 
 
+EVENT_KINDS = ['task', 'control']
+
+
 class Event(PRecord):
-    type = field(type=str,
-                 invariant=lambda x: x in ['task', 'control'])
+    kind = field(type=str,
+                 invariant=lambda x: (x in EVENT_KINDS,
+                                      'kind not in {}'.format(EVENT_KINDS)))
     # we store timestamps as seconds since epoch.
     # use time.time() to generate
     timestamp = field(type=float)
@@ -25,7 +29,7 @@ class Event(PRecord):
     # task_id this event pertains to
     task_id = field(type=str)
     # task config dict that sourced the task this event refers to
-    task_config = field(type=PMap)
+    task_config = field(type=dict)
     # the task finished with exit code 0
     success = field(type=(bool, type(None)), initial=None)
     # platform-specific event type
@@ -35,16 +39,14 @@ class Event(PRecord):
     message = field(type=str)
 
 
-class TaskEvent(Event):
-    def __init__(self, **kwargs):
-        kwargs.setdefault('type', 'task')
-        super(Event, self).__init__(**kwargs)
+def task_event(**kwargs):
+    kwargs.setdefault('kind', 'task')
+    return Event(**kwargs)
 
 
-class ControlEvent(Event):
-    def __init__(self, **kwargs):
-        kwargs.setdefault('type', 'control')
-        super(Event, self).__init__(**kwargs)
+def control_event(**kwargs):
+    kwargs.setdefault('kind', 'control')
+    return Event(**kwargs)
 
 
 def json_serializer(o):
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -9,20 +9,38 @@ from pyrsistent import PRecord
 
 
 class Event(PRecord):
+    type = field(type=str, initial=None)
     # we store timestamps as seconds since epoch.
     # use time.time() to generate
     timestamp = field(type=float)
     # reference to platform-specific event object
     raw = field()
-    # is this the last event for a task?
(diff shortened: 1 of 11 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -376,6 +377,17 @@ def _initialize_metrics(self):
...
     ####################################################################
 
+    def error(self, message):
+        event = ControlEvent.set(raw=message)
+
+        # TODO: have a mapper function similar to translator of task events
+        if message == 'Framework has been removed':
+            event = event.set(message='stop')
+        else:
+            event = event.set(message='unknown')
```
Rob-Johnson: I don't see ``message`` in the list of fields associated with the Event?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -376,6 +377,17 @@ def _initialize_metrics(self):
     #                   Mesos driver hooks go here                     #
     ####################################################################
 
+    def error(self, message):
+        event = ControlEvent.set(raw=message)
+
+        # TODO: have a mapper function similar to translator of task events
+        if message == 'Framework has been removed':
```
keymone: i totally agree, but i think we can iterate on this

task_processing/plugins/mesos/execution_framework.py
keymone: i must have lost it when i was shuffling code around, well spotted

task_processing/interfaces/event.py
```diff
@@ -8,18 +8,36 @@
 
 
 class Event(PRecord):
-    timestamp = field(type=datetime.datetime)
+    type = field(type=str, initial=None)
```
keymone: that was my original thinking, but considering that events are really just dicts and our type system is there purely for convenience, it is worth having the dict to be self-descriptive rather than relying on type being attached to it everywhere. we would lose that type information during serialization for instance.

keymone:
@sagar8192 that is true, but there will be many different events we would want to react to. rather than having a field per event type it's more pragmatic to have a single message field describing the event.

task_processing/interfaces/event.py
Rob-Johnson: you're right that we've not got a good system for serialization, but isn't that a problem to be solved at serialization time? i.e. set a type field in the record written to dynamo and then use that field to create an object of the correct type?

if you feel strongly, then please remove the class hierarchy all together, and encode the type only in the type field, rather than the type field and the class: there isn't anything to say I can't create a ControlEvent with a type 'foo', and confuses which encoding is the source of truth.

keymone: right, it is a problem that could be solved during serialization, but then you'd have to solve it multiple times in every place that reads our events, while having self-descriptive dict allows it to be interpreted in any environment with less hassle and side-steps the need to have special de/serialization procedures.

the reason i made these as separate types was to avoid having to provide timestamp every time we derive new event from existing one. we could replace those with `control_event` and `task_event` helper functions - would that be better?

keymone: also none of this is set in stone, point of this PR is to have a discussion and how to represent event hierarchy is very valid question. let's hear more opinions on this?

task_processing/runners/sync.py
```diff
@@ -20,6 +20,11 @@ def run(self, task_config):
 
         while True:
             event = event_queue.get()
+
+            if event.type == 'control' and \
```
Rob-Johnson: think these need to be {{event.kind}} now

task_processing/interfaces/event.py
```diff
@@ -5,24 +5,49 @@
...
+    # the task finished with exit code 0
+    success = field(type=(bool, type(None)), initial=None)
+    # platform-specific event type
+    platform_type = field(type=str)
+
+    # control events
+    message = field(type=str)
+
+
+def task_event(**kwargs):
```
jolynch: Why is this better than just having separate PRecords for TaskEvent and ControlEvent? I know inheritance hierarchies are sad but it's not that bad when you have type hierarchies imo.

keymone: having separate precords is totally fine, but it implies that we differentiate events by object class. if our events were only internal - i'd totally agree, doing `isinstance(e, ControlEvent)` is much better than doing `e.kind == 'control'`, however self-describing maps are trivially serializable, while to serialize type hierarchy we need to add one more layer and duplicate that layer at every place where our events are used (history browsing and whatnot).

if you think type hierarchy is still worth it - let's do it. but it's larger design decision and even though everybody gave me a shipit, I don't feel like we have consensus on self-describing maps. shall we schedule a quick call? cc/ @Rob-Johnson @sagar8192

PR #36, opened 2017-07-11
Add support for HA proxy

charleskwwan:
Summary of changes:
- add SmartStack plugin to get Mesos master info
- modified MesosExecutor to accept either a (cluster, region) or ip:port. Ideally, should be only the former, but need the latter for itests.
- update examples and tests to match
Testing:
- make test (with new SmartStack and modified MesosExecutor tests)
- examples excluding Promise - did not work previously, can't get it to work

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -14,6 +14,7 @@ from task_processing.plugins.mesos.execution_framework import (
     ExecutionFramework
 )
 from task_processing.plugins.mesos.translator import mesos_status_to_event
+from task_processing.plugins.smartstack import get_mesos_master
 FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s - %(message)s'
 LEVEL = logging.DEBUG
 logging.basicConfig(format=FORMAT, level=LEVEL)
@@ -56,7 +57,7 @@ class MesosExecutor(TaskExecutor):
         pool=None,
         principal='taskproc',
         secret=None,
-        mesos_address='127.0.0.1:5050',
+        mesos_info=('norcal_devc', 'uswest1-devc'),
         framework_translator=mesos_status_to_event,
         framework_name='taskproc-default',
         framework_staging_timeout=60,
@@ -78,7 +79,11 @@ class MesosExecutor(TaskExecutor):
             task_staging_timeout_s=framework_staging_timeout
         )
 
-        # TODO: Get mesos master ips from smartstack
+        if isinstance(mesos_info, tuple):
+            mesos_address = get_mesos_master(*mesos_info)
+        elif isinstance(mesos_info, str):
+            mesos_address = mesos_info
+
         self.driver = MesosSchedulerDriver(
             sched=self.execution_framework,
             framework=self.execution_framework.framework_info,
--- a/task_processing/plugins/smartstack.py
+++ b/task_processing/plugins/smartstack.py
@@ -1,0 +1,34 @@
+import yaml
+
+
+YOCALHOST_IP = '169.254.255.254'
+TASKPROC_SERVICE_FILE = '/nail/etc/services/task_processing/smartstack.yaml'
+
+
+class MesosLeaderNotFound(Exception):
+    pass
+
+
+class ProxyPortNotFound(Exception):
+    pass
+
+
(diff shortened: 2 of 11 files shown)


task_processing/plugins/smartstack.py
```diff
@@ -0,0 +1,34 @@
+import yaml
+
+
+YOCALHOST_IP = '169.254.255.254'
+TASKPROC_SERVICE_FILE = '/nail/etc/services/task_processing/smartstack.yaml'
```
keymone: i don't think these values should be anywhere in our repo. we should make library users provide them when initializing `MesosExecutor`.

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -14,6 +14,7 @@
     ExecutionFramework
 )
 from task_processing.plugins.mesos.translator import mesos_status_to_event
+from task_processing.plugins.smartstack import get_mesos_master
```
keymone: i would just make smartstack module part of mesos plugin. as implemented, `task_processing.plugins.smartstack` does nothing without mesos plugin and mesos plugin would fail to work without smartstack module due to this import.

charleskwwan: Fair.

task_processing/plugins/smartstack.py
charleskwwan: I agree for the services file, but the ip is link-local and useless to outside Yelp. I think it's okay?

sagar8192: +1

keymone: right, ip is just as useless to anybody outside yelp as the service file, which is exactly why i don't think it belongs in this repo. the fewer magic values we have hardcoded here - the better, lets make users provide all the inputs.

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -56,7 +57,7 @@ def __init__(
         pool=None,
         principal='taskproc',
         secret=None,
-        mesos_address='127.0.0.1:5050',
+        mesos_info=('norcal_devc', 'uswest1-devc'),
```
keymone: imo lets not default to something that doesn't make sense outside yelp

task_processing/plugins/mesos/mesos_executor.py
jolynch: I think that this is the wrong abstraction for this. The mesos executor should take just a connection ip, and the caller which is paasta has the smartstack information.

task_processing/plugins/smartstack.py
jolynch: This is not a public interface of smartstack
 You get /nail/etc/services.yaml and /var/run/synapse/services/<json discovery file>

task_processing/plugins/smartstack.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def get_mesos_master(cluster, region):
+    """ Finds the yocalhost ip:port for the Mesos master via SmartStack
+    Args customize granularity, defaults to a dev cluster for safety
+    """
+    with open(TASKPROC_SERVICE_FILE, 'r') as istream:
+        taskproc_conf = yaml.load(istream)
+
+    leader_name = 'mesos_leader_paasta_{cluster}_{region}'.format(
```
jolynch: This makes me think this belongs in paasta even more. External code bases can register plugins too, it doesn't have to all be in task proc :-)

task_processing/plugins/smartstack.py
charleskwwan: Gotcha.

task_processing/plugins/mesos/mesos_executor.py
charleskwwan: So basically, you're saying we just add a smartstack module that users can call before creating an executor.

task_processing/plugins/mesos/mesos_executor.py
charleskwwan: Fair.

charleskwwan:
Based on all the comments, I do agree that since adding haproxy support will need smartstack info that only paasta should have, I'll move the function to or create a plugin in paasta.

charleskwwan:
New request up at https://github.com/Yelp/paasta/pull/1368.

Will close this request now.

PR #37, opened 2017-07-11
blacklist agents with upcoming maintenance window

Rob-Johnson:
blacklist the agent until the maintenance window is completed.
this rearchitects each blacklisted slave to have a separate thread
responsible for taking the node out of the blacklist, rather than a single
blacklist thread responsible for managing all blacklisted agents.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -8,6 +8,7 @@ from pyrsistent import field
 from pyrsistent import m
 from pyrsistent import PRecord
 from pyrsistent import thaw
+from pyrsistent import v
 from six.moves.queue import Queue
 
 from task_processing.interfaces.event import control_event
@@ -87,7 +88,7 @@ class ExecutionFramework(Scheduler):
 
         self.offer_decline_filter = Dict(refuse_seconds=self.offer_backoff)
         self._lock = threading.RLock()
-        self.blacklisted_slaves = m()
+        self.blacklisted_slaves = v()
         self.task_metadata = m()
 
         self._initialize_metrics()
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -430,6 +423,28 @@ def resourceOffers(self, driver, offers):
...
+            now = int(time.time())
+            duration = completion_time - now
+            if not time.time() > completion_time:
+                self.blacklist_slave(
+                    agent_id=offer.agent_id.value,
+                    timeout=duration,
+                )
+
+        without_maintenance_window = [
+            offer for offer in offers if offer not in with_maintenance_window]
```
keymone: maybe move `without_maintenance_window` up to `with_maintenance_window` and use `not offer.unavailability` as condition? would read better to me. (also webscale!)

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -427,6 +420,28 @@ def resourceOffers(self, driver, offers):
...
+
+        for offer in with_maintenance_window:
+            start_time = offer.unavailability.start['nanoseconds']
+            completion_time = int(
+                (start_time + offer.unavailability.duration['nanoseconds'])
+                / 1000000000
+            )
+            now = int(time.time())
+            duration = completion_time - now
+            if not time.time() > completion_time:
```
keymone: duration > 0 # ? saves us one time.time() call

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -150,39 +148,34 @@ def offer_matches_pool(self, offer):
     def kill_task(self, task_id):
         self.driver.killTask(Dict(value=task_id))
 
-    def blacklist_slave(self, agent_id):
+    def blacklist_slave(self, agent_id, timeout):
         with self._lock:
```
keymone: blacklist_slave is only called from within kill_tasks_stuck_in_staging which already acquires the lock

keymone:
still shipit :ship:

task_processing/plugins/mesos/execution_framework.py
Rob-Johnson: nope - its also called from ``resourceOffers``

PR #38, opened 2017-07-12
fix volumes format confusion and add validator

keymone:


task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -20,6 +20,20 @@
...
+VOLUME_KEYS = set(['mode', 'container_path', 'host_path'])
+
+
+def valid_volumes(volumes):
+    for vol in volumes:
+        if vol.keys() != VOLUME_KEYS:
+            return (
+                False,
+                'Invalid volume format, must only contain following keys: '
+                '[mode, container_path, host_path], was: {}'.format(vol.keys())
```
Rob-Johnson: format this with ``VOLUME_KEYS``

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -20,6 +20,20 @@
 log = logging.getLogger(__name__)
 
 
+VOLUME_KEYS = set(['mode', 'container_path', 'host_path'])
```
Rob-Johnson: no need for this to be a set is there?

keymone: it seems like there is:

    In [3]: ['a', 'b'] == {'a': 1, 'b': 2}.keys()
    Out[3]: False

dict_keys supports equality comparison with sets but apparently not with lists

PR #40, opened 2017-07-12
aggregate offer outcomes and log once per resource offer

keymone:
instead of log line per offer on each resourceOffer call which is quite spammy when there are lots of agents

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -42,7 +42,6 @@ log = logging.getLogger(__name__)
 
 class TaskMetadata(PRecord):
     agent_id = field(type=str, initial='')
-    retries = field(type=int, initial=0)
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -470,13 +466,19 @@ def resourceOffers(self, driver, offers):
...
                             task_state='TASK_STAGING',
                             task_state_ts=time.time(),
                         )
                     )
                     get_metric(TASK_LAUNCHED_COUNT).count(1)
 
+        for reason, items in declined.items():
+            if items:
+                log.info(
+                    "Offers declined because {}: {}".format(reason, items))
```
sagar8192: Should we join the multiple offers?

keymone: that's kind of the point, otherwise we have line per offer which gets spammy fast

PR #42, opened 2017-07-17
Convert examples to use task processor + plugins

charleskwwan:
## Summary
- Converted examples to use taskproc plugins
- Moved stateful executor to plugins
## Testing done:
- Ran each example in docker before and after making changes
## Note to reviewers:
- Did not change promise example, since it does not seem to work, and promises have not been implemented anyway.


keymone:
shipit!

examples/async.py
```diff
@@ -22,10 +24,16 @@ def main():
...
     with open('./examples/cluster/secret') as f:
         secret = f.read().strip()
-    executor = MesosExecutor(
-        secret=secret,
-        mesos_address=mesos_address,
-        role='task-proc'
+
+    processor = TaskProcessor()
+    processor.load_plugin(provider_module='task_processing.plugins.mesos')
+    executor = processor.executor_from_config(
```
keymone: just wanted to comment that there are so many words in these 3 lines it would totally overwhelm me if i wanted to figure out how to use this stuff:

    processor.load_plugin(provider_module='task_processing.plugins.mesos')
    # processor, plugin, provider, module and executor on next line - that's five concepts

    # what if we made it look more like this:

    # registry better fits the functionality:
    taskproc_registry = TaskProcessingRegistry()

    # registry.load stuff from module is pretty self-explanatory:
    taskproc_registry.load(module='task_processing.plugins.mesos')

    # executor would be the only "new" concept so far, also lets just return the class:
    MesosExecutor = taskproc_registry.executor(provider='mesos')

    # pass args directly instead of through provider_arguments:
    executor = MesosExecutor(mesos_address=..., ...)

@sagar8192 @jolynch @Rob-Johnson thoughts?

PR #43, opened 2017-07-18
Added logic to deal with duplicate terminal status updates from mesos.

sagar8192:
Added code to handle duplicate status updates from mesos. Also fixed a parameter mismatch in reregistered method in execution framework.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -501,7 +501,12 @@ class ExecutionFramework(Scheduler):
             task=task_id
         ))
 
-        md = self.task_metadata[task_id]
+        try:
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -501,7 +501,13 @@ def statusUpdate(self, driver, update):
             task=task_id
         ))
 
+        if task_id not in self.task_metadata:
+            # We assume that a terminal status update has been
+            # received for this task already.
+            driver.acknowledgeStatusUpdate(update)
```
Rob-Johnson: can you log something here too, just to say we're ignoring the update?

sagar8192: Good call! Adding it.

PR #45, opened 2017-07-18
fix event task_config and retrying executor

keymone:
factory on task_config event field was coercing all PMap descendants to actual PMap instances, losing task_id property from MesosTaskConfig class

also minor fixes to locks and filtering of non-task events in retrying executor

### DIFF
--- a/task_processing/plugins/mesos/retrying_executor.py
+++ b/task_processing/plugins/mesos/retrying_executor.py
@@ -27,7 +27,6 @@ class RetryingExecutor(TaskExecutor):
         self.src_queue = executor.get_event_queue()
         self.dest_queue = Queue()
         self.stopping = False
-        self.TASK_CONFIG_INTERFACE = executor.TASK_CONFIG_INTERFACE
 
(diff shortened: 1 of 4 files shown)


task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -66,7 +65,7 @@ def retry_loop(self):
             while not self.src_queue.empty():
                 e = self.src_queue.get()
 
-                if e.kind == 'control':
+                if e.kind != 'task':
```
huadongliu: Are these two equivalent since EVENT_KINDS has only 'task' and 'control'?

keymone: they are, but we don't know what kind of events we might have in future and retrying executor is probably only ever concerned with task events

huadongliu: shipit

huadongliu:
shipit

Rob-Johnson:
shipit

PR #47, opened 2017-07-21
retrying executor and mesos framework fixes

keymone:
- make retrying executor more straightforward, counting retries from 0 to max
- log when max retries reached
- remove unnecessary sleep, due to thread not being daemon
- make retry loop thread daemon
- add metadata cleanup thread in mesos framework
- move setting agent_id in task metadata to where task is actually
  being launched
- remove discarding task metadata on first terminal event

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -82,7 +82,7 @@ class ExecutionFramework(Scheduler):
         self.event_queue = Queue(max_task_queue_size)
         self.driver = None
         self.are_offers_suppressed = False
-        self.suppress_after = int(time.time()) + suppress_delay
+        self.suppress_after = time.time() + suppress_delay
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -101,11 +101,22 @@ def __init__(
         }
 
         self.stopping = False
+
         task_kill_thread = threading.Thread(
             target=self.kill_tasks_stuck_in_staging, args=())
         task_kill_thread.daemon = True
         task_kill_thread.start()
 
+        # metadata is not removed immediately upon terminal event because mesos
```
sagar8192: If we receive a terminal status update for a task, then we remove it from the internal state immediately. I am not sure if this review was created before I added the fix for duplicate terminal events. Currently, we ignore duplicate status updates. Do we still need this review?

PR #48, opened 2017-07-25
[wip] changes to task config inheritance and new field

keymone:
- enforce task config inheritance from defaulttaskconfiginterface
- attach random uuid label to each new task
- build labels into extensions field on task config


keymone:
the reason we need this is that there is not enough information to handle multiple non-duplicate terminal events in retrying executor because task_id is constant throughout the stack.

execution framework will *always* attach a random uuid for every launched task which makes them distinguishable in upper levels.

it's a proof of concept and if you think there is better solution - lets discuss it here.

PR #49, opened 2017-08-02
Add gpu resource and mesos containerizer option

charleskwwan:
## Summary
To run GPU jobs via taskproc, we ideally should use mesos containerizer, which we currently do not support. As such, I have:
- Added gpus as a resource
- Added a containerizer option to MesosTaskConfig
- Changed the config sent to Mesos to use either mesos or docker containerizer
## Testing
- make test
- Sync run on norcal-devc to a gpu instance (role isolated); cmd runs a script that uses pytorch to check for gpu availability.
## Additional Notes
- (DONE) ~~Mesos 1.3.0 has a [bug](https://issues.apache.org/jira/browse/MESOS-7692) in which env vars are not properly set in docker containers, which causes any gpu scripts to not be able to see gpus. The bug has been resolved but its patch has not been released, so @nhandler is working on applying the patch directly.~~
- To make gpu runs fully functional, mesos-slaves will need to run with different options that make gpu resources available. That will take additional puppet/soa config changes, but is beyond the scope of this issue.
- Currently, my mesos containerizer option still enforces use of docker, though the api message is slightly different; making this option more flexible is trivial but beyond the scope of this issue.
- Mesos containerizer automatically makes GPUs available to docker containers (pymesos helped with this).
- Docker images run in mesos containerizer are responsible for ensuring all the gpu drivers are present in their containers.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -299,25 +299,34 @@ class ExecutionFramework(Scheduler):
                 md.set(agent_id=str(offer.agent_id.value))
             )
 
-        docker = Dict(
-            type='DOCKER',
-            docker=Dict(
-                image=task_config.image,
-                network='BRIDGE',
-                force_pull_image=True,
-                port_mappings=[Dict(host_port=port_to_use,
-                                    container_port=8888)]
-            ),
-            parameters=thaw(task_config.docker_parameters),
-            volumes=thaw(task_config.volumes),
-        )
         if task_config.containerizer == 'DOCKER':
-            container = docker
+            container = Dict(
+                type='DOCKER',
+                volumes=thaw(task_config.volumes),
+                docker=Dict(
+                    image=task_config.image,
+                    network='BRIDGE',
+                    port_mappings=[Dict(host_port=port_to_use,
+                                        container_port=8888)],
+                    parameters=thaw(task_config.docker_parameters),
+                    force_pull_image=True,
+                ),
+            )
         elif task_config.containerizer == 'MESOS':
-            docker.docker.name = docker.docker.pop('image')
             container = Dict(
                 type='MESOS',
-                mesos=Dict(image=docker),
+                # for docker, volumes should include parameters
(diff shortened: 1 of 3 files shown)


keymone:
nice! 🚢

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -295,6 +299,27 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+        docker = Dict(
+            type='DOCKER',
+            docker=Dict(
+                image=task_config.image,
+                network='BRIDGE',
+                force_pull_image=True,
+                port_mappings=[Dict(host_port=port_to_use,
+                                    container_port=8888)]
+            ),
+            parameters=thaw(task_config.docker_parameters),
```
solarkennedy: I'm surprised this gets used by the mesos containerizer?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -295,6 +299,27 @@ def create_new_docker_task(self, offer, task_config, available_ports):
                 md.set(agent_id=str(offer.agent_id.value))
             )
 
+        docker = Dict(
+            type='DOCKER',
+            docker=Dict(
+                image=task_config.image,
+                network='BRIDGE',
+                force_pull_image=True,
+                port_mappings=[Dict(host_port=port_to_use,
```
solarkennedy: I'm also skeptical this works in the mesos containerizer

task_processing/plugins/mesos/execution_framework.py
charleskwwan: I'll take another look at the protobufs and see if I need to make any changes.

charleskwwan:
Looked at the [Mesos protobuf documentation](https://github.com/apache/mesos/blob/c65fb009e452e9a6cb0c9e2f8541423b74dd64cc/include/mesos/mesos.proto) and made the appropriate changes.

Also, rolled back the DRY to make it easier to read as per @solarkennedy's suggestion.

PR #50, opened 2017-08-02
Add mesos master smartstack discovery as util

charleskwwan:
## Summary
- Added a util func to get mesos master (or any cluster master) from a services file
## Testing
- make test
- Manually run the func to get the address
## Notes
- Tried to exclude as much Yelp-y info as possible, but the default is still our own services file location


huadongliu:
shipit

PR #55, opened 2017-08-15
Deal with TASK_LOST due to invalid offers

sagar8192:
I think it is better to re-enqueue the task in execution_framework than doing it in retrying_executor because not all applications have to use retrying_executor. This is something that execution framework should be able to deal with.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -545,8 +545,8 @@ class ExecutionFramework(Scheduler):
         if task_id not in self.task_metadata:
             # We assume that a terminal status update has been
             # received for this task already.
-            log.info('Ignoring this status update because a terminal status'
-                     'update has been receiced for this task already.')
+            log.info('Ignoring this status update because a terminal status '
+                     'update has been received for this task already.')
             driver.acknowledgeStatusUpdate(update)
             return
 
@@ -561,8 +561,9 @@ class ExecutionFramework(Scheduler):
             # This task has not been launched. Therefore, we are going to
             # reenqueue it. We are not propogating any event up to the
             # application.
-            log.warning('Received TASK_LOST from mesos master because we'
-                        'attemted to accept an invalid offer.')
+            log.warning('Received TASK_LOST from mesos master because we '
+                        'attempted to accept an invalid offer. Going to re-'
+                        'enqueue this task {id}'.format(id=task_id))
             self.task_metadata = self.task_metadata.discard(task_id)
             self.enqueue_task(md.task_config)
             get_metric(TASK_LOST_DUE_TO_INVALID_OFFER_COUNT).count(1)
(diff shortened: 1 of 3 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -542,12 +545,30 @@ def statusUpdate(self, driver, update):
...
+        # mesos master.
+        if task_state == 'TASK_LOST' and \
+                'REASON_INVALID_OFFERS' == str(update.reason):
+            # This task has not been launched. Therefore, we are going to
+            # reenqueue it. We are not propogating any event up to the
+            # application.
+            log.warning('Received TASK_LOST from mesos master because we'
+                        'attemted to accept an invalid offer.')
+            self.task_metadata = self.task_metadata.discard(task_id)
+            self.enqueue_task(md.task_config)
```
keymone: i'd prefer this to be a separate retrying layer with

    retry_pred = lambda x: str(x.state) == 'TASK_LOST' and str(x.reason) == 'REASON_INVALID_OFFERS'

reasons:

- less code in mesos framework
- library users have better control over the stack
- incoming task_id refactoring is going to make this code tricky - we shouldn't enqueue task with same id but we also shouldn't be mutating task_config.task_id because it won't be recognizable up the stack

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -542,12 +545,30 @@ def statusUpdate(self, driver, update):
         if task_id not in self.task_metadata:
             # We assume that a terminal status update has been
             # received for this task already.
-            log.info('Ignoring this status update because a terminal status \
-                update has been receiced for this task already.')
+            log.info('Ignoring this status update because a terminal status'
+                     'update has been receiced for this task already.')
```
Rob-Johnson: can you fix the typo while you're here please?

task_processing/plugins/mesos/execution_framework.py
Rob-Johnson: I think I agree - lets be consistent with our use of the executor and not make exceptions. We can make a 'default_mesos_stack' that includes all those executors we deem to be sensible - there is no reason to force everyone to do this.

huadongliu: I am not sure the library (RetryingExecutor) users need/want to know TASK_LOST and REASON_INVALID_OFFERS. They are infrastructure details. It is the framework's responsibility to deal with such to provide a robust processing infrastructure. TASK_LOST with REASON_INVALID_OFFERS is different from other failures because we know the task hasn't been touched and it is safe to retry.

sagar8192: I decided to put the retry logic in execution_framework because this does not count as a retry. This task has not been even launched. This issue is due to infra/mesos issues which our framework should be able to deal with.

@keymone when we re-enqueue a task, we just change the metadata which is only used by execution_framework. Applications only care about the terminal state of the task.

keymone: i agree that it's infrastructure details and we shouldn't concern users with these issues. what i'd like to avoid is large classes with many responsibilities. execution framework is already 600 lines long and we still have more mesos-specific things to add (reconciliations, etc).

my thinking is:

- executors are very isolated pieces of logic, it's easy to merge them into the framework if we later realize that some layer is an overkill
- on the other hand if some logic is embedded into the framework - over time it will be increasingly complex to rip it out and present in form of executor, execution framework will just snowball out of control

i'd rather do the easy thing early than potentially deal with hard thing later.

PR #56, opened 2017-08-15
Mutable task id

keymone:
- move task_id out of task config
- task_id is collection of strings
- allow stack layers to extend/shrink task_id as it travels the stack with task runs and events
- use base62 encoded ids instead of much more verbose uuids, underlying data is the same - random 16 bytes

### DIFF
--- a/examples/dynamo_persistence.py
+++ b/examples/dynamo_persistence.py
@@ -59,15 +59,15 @@ def main():
         }
     )
     runner = Sync(executor=executor)
-    tasks = set()
     TaskConfig = mesos_executor.TASK_CONFIG_INTERFACE
-    for _ in range(1, 2):
-        task_config = TaskConfig(
-            image='ubuntu:14.04', cmd='/bin/sleep 2'
-        )
-        tasks.add(task_config.task_id)
-        runner.run(task_config)
-        print(executor.status(task_config.task_id))
+
+    for _ in range(2):
+        task_config = TaskConfig(image='ubuntu:14.04', cmd='/bin/true')
+        task_id = runner.new_task_id()
+        runner.run(task_config, task_id)
+        assert executor.status(task_id)[-1]['terminal']
+
+    print('OK')
 
 
 def create_table(client):
--- a/examples/dynamo_persistence.py
+++ b/examples/dynamo_persistence.py
@@ -5,6 +5,7 @@ import os
 from boto3 import session
 from botocore.errorfactory import ClientError
 
+from task_processing.interfaces.runner import new_task_id
 from task_processing.plugins.persistence.dynamodb_persistence \
     import DynamoDBPersister
 from task_processing.runners.sync import Sync
@@ -63,7 +64,7 @@ def main():
 
     for _ in range(2):
         task_config = TaskConfig(image='ubuntu:14.04', cmd='/bin/true')
-        task_id = runner.new_task_id()
+        task_id = new_task_id()
         runner.run(task_config, task_id)
         assert executor.status(task_id)[-1]['terminal']
 
--- a/task_processing/interfaces/runner.py
+++ b/task_processing/interfaces/runner.py
@@ -1,10 +1,10 @@
 import abc
-import uuid
 
 import six
 from pyrsistent import pvec
 
 from task_processing.interfaces.task_executor import DefaultTaskConfigInterface
+from task_processing.utils.base62 import base62_random
 
 
 @six.add_metaclass(abc.ABCMeta)
@@ -24,4 +24,4 @@ class Runner(object):
         pass
 
     def new_task_id():
-        return pvec(uuid.uuid4())
+        return pvec(base62_random())
--- a/task_processing/interfaces/runner.py
+++ b/task_processing/interfaces/runner.py
@@ -1,10 +1,10 @@
 import abc
 
 import six
-from pyrsistent import pvec
+from pyrsistent import pvector
 
 from task_processing.interfaces.task_executor import DefaultTaskConfigInterface
-from task_processing.utils.base62 import base62_random
+from task_processing.util.base62 import base62_random
 
 
 @six.add_metaclass(abc.ABCMeta)
@@ -24,4 +24,4 @@ class Runner(object):
         pass
 
     def new_task_id():
-        return pvec(base62_random())
+        return pvector(base62_random())
--- a/task_processing/interfaces/runner.py
+++ b/task_processing/interfaces/runner.py
@@ -23,5 +23,5 @@ class Runner(object):
     def kill(self, task_id):
         pass
 
-    def new_task_id():
-        return pvector(base62_random())
+    def new_task_id(self):
+        return pvector([base62_random()])
--- a/task_processing/interfaces/runner.py
+++ b/task_processing/interfaces/runner.py
@@ -1,7 +1,6 @@
 import abc
 
 import six
-from pyrsistent import pvector
 
 from task_processing.interfaces.task_executor import DefaultTaskConfigInterface
 from task_processing.util.base62 import base62_random
@@ -23,5 +22,6 @@ class Runner(object):
     def kill(self, task_id):
         pass
 
-    def new_task_id(self):
-        return pvector([base62_random()])
+
+def new_task_id():
+    return base62_random()
--- a/task_processing/interfaces/runner.py
+++ b/task_processing/interfaces/runner.py
@@ -1,6 +1,8 @@
 import abc
+import uuid
 
 import six
+from pyrsistent import pvec
 
 from task_processing.interfaces.task_executor import DefaultTaskConfigInterface
 
@@ -14,9 +16,12 @@ class Runner(object):
     """
 
     @abc.abstractmethod
-    def run(self, task_config):
+    def run(self, task_config, task_id=None):
         pass
 
     @abc.abstractmethod
     def kill(self, task_id):
         pass
+
+    def new_task_id():
+        return pvec(uuid.uuid4())
--- a/task_processing/interfaces/task_executor.py
+++ b/task_processing/interfaces/task_executor.py
@@ -28,7 +28,7 @@ class TaskExecutor(object):
         The executor should start running the provided task and return the
         task id.
 
-        :returns pvector task_id: Callers get the id of the task that was run
+        :returns str task_id: Callers get the id of the task that was run
         to check status or kill it later
         """
         pass
@@ -51,6 +51,6 @@ class TaskExecutor(object):
     def get_event_queue(self):
         """Get queue of events
 
-        :returns: Object with .pull and .push methods defined on it.
+        :returns: Object with .get and .put methods defined on it.
         """
         pass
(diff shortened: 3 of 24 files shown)


task_processing/util/base62.py
```diff
@@ -0,0 +1,30 @@
+import os
```
Rob-Johnson: why do we need to do all this? whats wrong with UUIDs?

keymone: they are very verbose. if we have task id that contains more than single uuid (which isn't impossible) logs will be unreadable. keep in mind, the data that's base62-encoded is identical to uuid bytes so we're not compromising anything here.

keymone:
@robj you're right. we'll have a problem either way though - if a task_id constructed by stack X is attempted to be parsed by stack Y it won't matter if we retain layer information of each component - stack Y just won't know what to do with it.

not entirely sure how to go about it. giving each id-mutating layer some unique-ish name and storing that along each component in task_id will bloat the ids a lot and even base62 won't save us. but it is useful info to store.

also it might not matter in the end - so far there is only one executor that needs this functionality - retrying, and it's id components won't be uuids, it'll be 1..n - short enough to fit in the id string?

keymone:
@sagar8192 sure, it's in TASKPROC-89

keymone:
rebased off master

sagar8192:
@keymone  In execution framework, we remove the task_id from the internal state after getting an update with a terminal state(Ref: https://github.com/Yelp/task_processing/blob/master/task_processing/plugins/mesos/execution_framework.py#L545). We also send an event to upper layers. Next time when we get a duplicate update, we simply discard it. This should prevent us from propagating duplicate status updates for the same task, right? Am I missing something?

keymone:
@sagar8192 right, but after task has failed, retrying executor will restart it with the same task id and duplicate terminal event will apply to the second attempt of the same task, causing it to look as if it failed as well.

sagar8192:
Ah, right. So, for each retry, if we can add append retry number(<task-id>-retry<no>), will it solve the issue? We don't have to keep track of all the retries.

keymone:
@sagar8192 yep, that's what this PR is about, just in more generic form, so that all executors know the mechanics of/guarantees about of task id mutations.

sagar8192:
This change looks good to me. The only outstanding question I have is - if we use retry<number> as the suffix for retry, then it would reduce the complexity and would also make it easier to debug.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -253,18 +259,18 @@ def get_tasks_to_launch(self, offer):
         with self._lock:
             # Get all the tasks of the queue
             while not self.task_queue.empty():
-                task = self.task_queue.get()
+                task, mesos_task_id = self.task_queue.get()
```
huadongliu: Please rename task to task_config.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -253,18 +259,18 @@ def get_tasks_to_launch(self, offer):
...
                     # This offer is sufficient for us to launch task
                     tasks_to_launch.append(
-                        self.create_new_docker_task(
-                            offer,
-                            task,
-                            available_ports
+                        (
+                            self.create_new_docker_task(
+                                offer, task, mesos_task_id, available_ports),
+                            mesos_task_id
```
huadongliu: mesos_task_id is the task_id field of TaskInfo. I don't see a need for (task_id, taskInfo) tuple instead of just TaskInfo.

task_processing/interfaces/runner.py
```diff
@@ -14,9 +16,12 @@ class Runner(object):
...
     @abc.abstractmethod
-    def run(self, task_config):
+    def run(self, task_config, task_id=None):
         pass
 
     @abc.abstractmethod
     def kill(self, task_id):
         pass
+
+    def new_task_id(self):
```
sagar8192: I don't think this should be part of the interface. base62_random() is util method that people can just import and use.

task_processing/interfaces/task_executor.py
```diff
@@ -24,14 +21,14 @@ class TaskExecutor(object):
     """
 
     @abc.abstractmethod
-    def run(self, task_config):
-        """Run the supplied task
+    def run(self, task_config, task_id):
```
sagar8192: hmm... I don't understand why we need to pass task_id separately? Also, why are we using a vector?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -178,19 +181,22 @@ def unblacklist_slave(self, agent_id, timeout):
             self.blacklisted_slaves = \
                 self.blacklisted_slaves.remove(agent_id)
 
-    def enqueue_task(self, task_config):
+    def enqueue_task(self, task_config, task_id):
+        mesos_task_id = ':'.join(task_id)
```
sagar8192: I am a bit confused about the need for mesos_task_id and task_id. Why do we need mesos_task_id? IMO, it is complicating the code a lot.

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -87,12 +96,13 @@ def retry_loop(self):
...
-        if task_config.task_id not in self.task_retries:
+    def run(self, task_config, task_id):
+        if task_id not in self.task_retries:
             with self.task_retries_lock:
                 self.task_retries = self.task_retries.set(
-                    task_config.task_id, self.retries)
-        self.executor.run(task_config)
+                    task_id, self.retries)
+
+        self.executor.run(task_config, self.grow(task_id))
```
huadongliu: What if a user calls run() with the same task_id? It results in one retry without task_retries map being updated. More importantly, locking is needed to avoid races with the retry loop.

task_processing/plugins/mesos/execution_framework.py
keymone: mesos_task_id is the actual value that goes to mesos over a wire. task_id is what lives in our stack. execution_framework handles mapping one to another and making sure neither leaks where it doens't belong.

task_processing/plugins/mesos/retrying_executor.py
keymone: default interface of the runner has optional task_id, so if user passes the id that has already been used - they supposedly know what they are doing. the effect is the same as if user had called .run with the identical task_config using the current master branch.

but it's a good point, this is undocumented and undefined behavior. we should decide what do we really want to happen when user does this. do we want to force another retry if the task is still running? do we want to restart from 0? maybe only if task had already finished?

task_processing/interfaces/task_executor.py
keymone: we need to pass it separately to make sure task_config is not mutated on the way and it's a vector to reduce the assumptions every executor has to make and share about format of task id, how it's mutated in the stack, how to break it up into components, etc. by having task_id a vector all other executors need to know is that they can grow/shrink it on the tail, that's it.

task_processing/interfaces/runner.py
keymone: task_id is something we don't want users to touch at all. but if they do - let's give them a function that makes it easy.

examples/dynamo_persistence.py
```diff
@@ -59,15 +59,15 @@ def main():
...
-        task_config = TaskConfig(
-            image='ubuntu:14.04', cmd='/bin/sleep 2'
-        )
-        tasks.add(task_config.task_id)
-        runner.run(task_config)
-        print(executor.status(task_config.task_id))
+
+    for _ in range(2):
+        task_config = TaskConfig(image='ubuntu:14.04', cmd='/bin/true')
+        task_id = runner.new_task_id()
```
huadongliu: I think the case to pass in a task_id instead of relying on the runner to generate one is when a task_id is needed later, for example, to check status or terminate its execution. I would recommend not using new_task_id in examples unless it is used later.

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -108,3 +118,9 @@ def stop(self):
 
     def get_event_queue(self):
         return self.dest_queue
+
+    def grow(self, task_id):
+        return task_id.append(self.retries - self.task_retries[task_id])
+
+    def shrink(self, task_id):
+        return task_id[0:-1]
```
huadongliu: I am not sure about assuming a special format of task_id (e.g. X:Y:Z) is a robust way to deal with delayed/duplicated updates. Users now have a choice to pass in task_id without control. ':' may be included in a user generated task_id. The bullet proof way is to keep track of retried task_id, more work, but safer.

task_processing/interfaces/task_executor.py
sagar8192: I think this is where the disagreement is. I am advocating for a solution that only involves changing the retrying executor. 
Something like this: https://gist.github.com/sagar8192/bb98d0b479b3cf1afcf09785529471fd

This way the translation logic lives only in one layer. Other layers do not need to know anything about details of the retrying executor(do not need to join task_id etc). We can add logging to make sure that task_id mapping info is captured for debugging. IMO, this change is less intrusive and also does not introduce complexity in other layers.

task_processing/plugins/mesos/retrying_executor.py
sagar8192: I am of the opinion that we should not do any validation. If people an enqueuing tasks with duplicate ids, then the application should just crash. We should not make any assumptions on their behalf.

task_processing/plugins/mesos/retrying_executor.py
sagar8192: This is a good point!

keymone: enforcing a "no colon in id components" rule is fairly easy, tracking and mapping new ids for each retry is much harder and context is completely lost for every layer below retrying.

task_processing/interfaces/task_executor.py
keymone: the problem with this solution is that

- we're mutating supposedly immutable task_config (other layers can depend on this)

- retrying executor itself depends on not mutating task_config and there's already a bug in your patch where after second retry on line 65 you'll "restore" not the original task_id but one from the first retry, that you store by mutating task_config in .run() (you restore event.task_id but not event.task_config.task_id)

- we need to maintain id mappings which is lost upon crashing so it makes recovery that much harder (all in-flight tasks must be killed unless original task_id and number of attempts is stored somewhere along them)

- task id mutation is something many layers may choose to implement and it's better to have it done one way than to have to keep track of multiple implementations when debugging, introspecting or implementing layers that have to deal with id components

all of these are addressed by separating task_id out and defining a clear way to add/remove components to it. yes, it increases complexity of everything by a little bit but removes the spikes in complexity of having to deal with all these problems in individual layers.

keymone:
@sagar8192 @Rob-Johnson @huadongliu i've changed task_id to be string and rebase off master, please give it another look and let me know if it's shippable.

huadongliu:
Mesos framework and retrying executor changes look good to me.

While working on the per-task time out executor, I feel that we will grow task config for new features. Some of the task config fields might be mutable. For those fields, are we going to change the API to separate them from fields that do not change? Can we make a copy of task config with an updated field and keep the API unchanged?

PR #60, opened 2017-08-23
Use official Mesosphere Mesos image for quicker builds.

ssk2:
It looks like we're not doing anything fancy here, so in the interest of speeding up our build times (specifically for running the integration tests), I changed the Dockerfile to use the upstream Docker image as a base for our testing image. This has the side effect of making the Dockerfile simpler too.

We could probably get away with using the upstream Docker image directly too if we don't want to set up secrets (is this a hard requirement?)

[Current builds](https://travis-ci.org/Yelp/task_processing/builds) take on the order of ~ 9 minutes. [This change](https://travis-ci.org/Yelp/task_processing/builds/267400488?utm_source=github_status&utm_medium=notification) will shorten it to ~ 7.5 minutes.


keymone:
🚢

examples/cluster/Dockerfile.mesos
```diff
@@ -12,17 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:trusty
+FROM mesosphere/mesos:1.2.0
```
sagar8192: Rather than building this image, can we just specify mesosphere/mesos:1.2.0 in docker-compose file? We can mount mesos-secrets and mesos-slave-secrets to the container using volumes option in docker-compose.

ssk2: Oh yes, good call. That'll actually save us even more time :).

huadongliu: I made a comment to use the 1.3.0 tag in the other PR.

ssk2: Done!

PR #62, opened 2017-08-24
Discard offers with 0 ports

sagar8192:
Currently, we are not honoring ports field in task_config. Users can pass more than one port to map, but we are mapping 8888 internal port to one of the available ports. This can be fixed in another review.


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -585,8 +586,8 @@ def statusUpdate(self, driver, update):
                     )
                 )
 
-        if task_state in self._task_states:
-            with self._lock:
+        with self._lock:
```
Rob-Johnson: why're you swapping these round?

PR #63, opened 2017-08-25
support setting environment vars

Rob-Johnson:
internal TASKPROC-107

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -6,6 +6,7 @@ from pymesos import MesosSchedulerDriver
 from pyrsistent import field
 from pyrsistent import m
 from pyrsistent import PMap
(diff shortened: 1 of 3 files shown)


task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -70,6 +72,7 @@ class MesosTaskConfig(PRecord):
                           invariant=lambda c:
                           (c == 'DOCKER' or c == 'MESOS',
                            'containerizer is docker or mesos'))
+    environment = field(type=PMap, initial=m())
```
keymone: add `factory=pmap` ?

huadongliu:
Shipit. It would be nice to have an example task config like 

task_config = TaskConfig(image="busybox", cmd='echo $TESTENV', environment={"TESTENV": "taskproc_sync"})

PR #64, opened 2017-08-25
Keep track of task state history instead of the last state update

huadongliu:
This change replaces the last state update timestamp with a {state: timestamp} map to keep track of task state transitions until a terminal state. 

I discarded a change earlier to enforce task execution timeout based on states in favor of a timeout executor. Still, expanding the last state to a set of states is useful in debugging because the stateful executor is not always used by all applications.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -6,6 +6,8 @@ from addict import Dict
 from pymesos.interface import Scheduler
 from pyrsistent import field
 from pyrsistent import m
+from pyrsistent import PMap
+from pyrsistent import pmap
 from pyrsistent import PRecord
 from pyrsistent import thaw
 from pyrsistent import v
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -180,12 +182,14 @@ def unblacklist_slave(self, agent_id, timeout):
...
         with self._lock:
+            # task_state and task_state_history get reset every time
+            # a task is enqueued.
             self.task_metadata = self.task_metadata.set(
                 task_config.task_id,
                 TaskMetadata(
                     task_config=task_config,
                     task_state='TASK_INITED',
-                    task_state_ts=time.time(),
+                    task_state_history=m().set('TASK_INITED', time.time()),
```
keymone: here and in other places you can just write `m(TASK_INITED=time.time())`

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -46,7 +47,7 @@ class TaskMetadata(PRecord):
     agent_id = field(type=str, initial='')
     task_config = field(type=PRecord, mandatory=True)
     task_state = field(type=str, mandatory=True)
-    task_state_ts = field(type=float, mandatory=True)
+    task_state_history = field(type=PMap, mandatory=True)
```
keymone: for convenience i would add `factory=pmap` - it'll automatically convert python dicts into pmaps on assignment

huadongliu:
I am not sure if A->B->A is possible. I would rather assert if not.

keymone:
@huadongliu yeah, if it's something we're sure about - shipit.

PR #66, opened 2017-08-30
TimeoutExecutor

huadongliu:
I had to prefix the example test command sleep with dumb-init. Otherwise, TASK_KILLED took 120 seconds because SIGTERM was not delivered to the sleep process and we use 120 seconds as mesos agent docker_stop_timeout.

### DIFF
--- a/examples/timeout.py
+++ b/examples/timeout.py
@@ -15,7 +15,7 @@ def main():
     args = parse_args()
     processor = TaskProcessor()
     processor.load_plugin(provider_module='task_processing.plugins.mesos')
-    executor = processor.executor_from_config(
+    mesos_executor = processor.executor_from_config(
         provider='mesos',
         provider_config={
             'secret': args.secret,
@@ -25,18 +25,24 @@ def main():
         }
     )
 
-    TaskConfig = executor.TASK_CONFIG_INTERFACE
-    task_config = TaskConfig(image="busybox", cmd='/bin/true')
-    # This only works on agents that have added mesos as a containerizer
-    # task_config = TaskConfig(containerizer='MESOS', cmd='/bin/true')
+    executor = processor.executor_from_config(
+        provider='timeout',
+        provider_config={
+            'downstream_executor': mesos_executor,
+        }
+    )
 
-    runner = Sync(executor)
+    TaskConfig = mesos_executor.TASK_CONFIG_INTERFACE
+    runner = Sync(executor=executor)
+    task_config = TaskConfig(
+        image='docker-dev.yelpcorp.com/dumb-busybox',
+        cmd='exec dumb-init /bin/sleep 30',
+        timeout=15
+    )
     result = runner.run(task_config)
     print(result)
-    print(result.raw)
-    runner.stop()
 
-    return 0 if result.success else 1
+    runner.stop()
 
 
 if __name__ == "__main__":
--- a/task_processing/plugins/mesos/timeout_executor.py
+++ b/task_processing/plugins/mesos/timeout_executor.py
@@ -1,0 +1,111 @@
+import collections
+import logging
+import time
+from threading import Lock
+from threading import Thread
+
+from six.moves.queue import Queue
+
+from task_processing.interfaces.task_executor import TaskExecutor
+
+log = logging.getLogger(__name__)
+
+TaskEntry = collections.namedtuple('TaskEntry', ['task_id', 'deadline'])
+
+
+class TimeoutExecutor(TaskExecutor):
+    def __init__(self, downstream_executor):
+        self.downstream_executor = downstream_executor
+
+        self.tasks_lock = Lock()
+        # Tasks that are pending termination
+        self.killed_tasks = []
+        # Tasks that are currently running
+        self.running_tasks = []
+
+        self.src_queue = downstream_executor.get_event_queue()
+        self.dest_queue = Queue()
+        self.stopping = False
+
+        self.timeout_thread = Thread(target=self.timeout_loop)
+        self.timeout_thread.daemon = True
+        self.timeout_thread.start()
+
+    def timeout_loop(self):
+        while True:
+            # process downstream events
+            while not self.src_queue.empty():
+                e = self.src_queue.get()
+                self.dest_queue.put(e)
+
+                if not (e.kind == 'task' and e.terminal):
+                    continue
+                # Update running and killed tasks
+                with self.tasks_lock:
+                    for idx, entry in enumerate(self.running_tasks):
+                        if e.task_id == entry.task_id:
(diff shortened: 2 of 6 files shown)


requirements.txt
```diff
@@ -1,4 +1,5 @@
 # Library so -e .
 -e .
 boto3
+cllist
```
keymone: i'm a bit reluctant to introducing new dependencies, can we get by with native python lists/maps for now and fix performance issues when they actually arise? considering how well we're doing on code isolation - migrating executor to different data structures won't be that complicated.

examples/timeout.py
```diff
@@ -0,0 +1,90 @@
...
+
+from task_processing.runners.sync import Sync
+from task_processing.task_processor import TaskProcessor
+
+FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s - %(message)s'
+LEVEL = logging.DEBUG
+logging.basicConfig(format=FORMAT, level=LEVEL)
+
+
+def parse_sync_args():
```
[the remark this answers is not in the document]
keymone: this is not actually request to change, i'm just thinking out loud, ship this bit, i'll make separate PR for extracting common bits from examples.

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,110 @@
...
+log = logging.getLogger(__name__)
+
+
+class TimeoutExecutor(TaskExecutor):
+    def __init__(self, downstream_executor, timeout_s):
+        self.downstream_executor = downstream_executor
+        if timeout_s <= 0:
+            raise ValueError('TimeoutExecutor expects a positive timeout')
+        else:
+            self.timeout_s = timeout_s
```
Rob-Johnson: I wonder if we should make the timeout a per task thing, rather than something that exists on the executor level? I think that might give more flexibility

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # Tasks that are pending termination
+        self.killed_tasks = []
+
+        # Tasks that are currently running
+        self.flying_tasks = dllist()
```
Rob-Johnson: can you call this running_tasks instead?

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,110 @@
...
+                            self.flying_tasks_map.pop(e.task_id)
+                    # remove task from killed list
+                    elif e.task_id in self.killed_tasks:
+                        self.killed_tasks.remove(e.task_id)
+
+            # Check timeouts
+            current_time = time.time()
+            with self.flying_tasks_lock:
+                dllistnode = self.flying_tasks.first
+                while dllistnode:
```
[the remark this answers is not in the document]
huadongliu: The reason to use a doubly linked list (ordered by timeout) with a map was for fast removal from the running list. I will drop the cllist to simplify things.

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,109 @@
...
+            self.running_tasks.append(new_entry)
+
+    def kill(self, task_id):
+        with self.tasks_lock:
+            for idx, entry in enumerate(self.running_tasks):
+                if task_id == entry.task_id:
+                    self.running_tasks.pop(idx)
+                    self.killed_tasks.append(task_id)
+                    log.info('Killing task {}: requested'.format(task_id))
+                    return
```
keymone: shouldn't we call `self.downstream_executor.kill` in here?

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,109 @@
...
+TaskEntry = collections.namedtuple('TaskEntry', ['task_id', 'deadline'])
+
+
+class TimeoutExecutor(TaskExecutor):
+    def __init__(self, downstream_executor):
+        self.downstream_executor = downstream_executor
+
+        self.tasks_lock = Lock()
+        # Tasks that are pending termination
+        self.killed_tasks = []
```
keymone: not sure i understand the purpose of this collection. in the code we walk running_tasks to invoke downstream_executor.kill, i don't see killed_tasks entries used for anything?

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,109 @@
...
+            # Since src_queue has to be polled continuously, sleep(1) is used.
+            # Otherwise, a notify() from run() plus wait(delta between now and
+            # the earliest deadline) is more efficient.
+            time.sleep(1)
+
+    def run(self, task_config):
+        # Tasks are dynamically added and removed from running_tasks and
+        # and killed_tasks. It's preferable for the client or execution
+        # framework to check for duplicated tasks.
+        self.downstream_executor.run(task_config)
```
keymone: i'd prefer this to be the last call in a function to avoid races with another thread:

- run sends task `ABC` to mesos, which fails immediately
- few lines later `with self.tasks_lock` waits for lock to be released from `timeout_loop` processing events for `XYZ`
- event queue gets terminal event for `ABC` before queue gets exhausted
- ???
- run appends already finished `ABC` to `running_tasks`

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -0,0 +1,109 @@
...
+        new_entry = TaskEntry(
+            task_id=task_config.task_id,
+            deadline=task_config.timeout + time.time()
+        )
+        with self.tasks_lock:
+            for idx, entry in enumerate(self.running_tasks):
```
keymone: not sure this is needed as `running_tasks.append` only happens in same thread in this method so `time.time()` is almost-guaranteed to be non-decreasing

task_processing/plugins/mesos/timeout_executor.py
huadongliu: Having killed_tasks or not won't affect functionality. It is for tracking/debug purposes. A terminal event is expected for a task being killed.

task_processing/plugins/mesos/timeout_executor.py
huadongliu: Fixed.

task_processing/plugins/mesos/timeout_executor.py
huadongliu: time.time() is non-decreasing. However, running_tasks is sorted by (enqueue_time+timeout).

task_processing/plugins/mesos/timeout_executor.py
huadongliu: Good catch!

PR #68, opened 2017-09-01
Support mesos containerizer without an image

huadongliu:
This is TASKPROC-78. Mesos containerizer with Docker image support has been completed for the GPU work. This is to support running a command directly in a mesos containerizer.

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -38,9 +38,14 @@ def valid_volumes(volumes):
 
 
 class MesosTaskConfig(PRecord):
+    def __invariant__(conf): return ('image' in conf if conf.containerizer ==
+                                     'DOCKER' else True,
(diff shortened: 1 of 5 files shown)


task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -40,7 +40,8 @@ def valid_volumes(volumes):
 class MesosTaskConfig(PRecord):
     uuid = field(type=uuid.UUID, initial=uuid.uuid4)
     name = field(type=str, initial="default")
-    image = field(type=str, initial="ubuntu:xenial")
+    # image is optional for the mesos containerizer
```
Rob-Johnson: can we add an invariant here to enforce this?

Reading the docs, we can have something like:

```
class MesosTaskConfig(PRecord):
    __invariant__ = lambda config: (config.image is not None if config.containerizer is not 'MESOS', 'Image required for chosen containerizer')
    ....
```

PR #69, opened 2017-09-18
Deal with launchTasks failures.

sagar8192:
This PR is to deal with launchTasks failures. 
If launchTask() fails, then there 2 outcomes:
1) Mesos has received the task.
2) Mesos did not receive the task.

The current approach is to move the tasks into unknown state and reconcile their status periodically. If the tasks have been received by mesos, then it will return their current state. But, they were never received by mesos, then it is our responsibility to re-enqueue these tasks.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -118,12 +118,37 @@ class ExecutionFramework(Scheduler):
             with self._lock:
                 time_now = time.time()
                 for task_id in self.task_metadata.keys():
-                    if self.task_metadata[task_id].task_state not in (
+                    eligible_for_termination_at = self.task_metadata[task_id].\
+                        task_state_ts + self.task_staging_timeout_s
+
+                    md = self.task_metadata[task_id]
+
+                    if time_now < eligible_for_termination_at:
+                        return
+
+                    if md.task_state == 'UNKNOWN':
+                        with self._lock:
+                            self.task_metadata = self.task_metadata.discard(
+                                task_id
+                            )
+                            self.enqueue_task(md.task_config)
+
+                            print('Would have killed and'
+                                  're-enqueued the task')
+                            continue
+
+                    if md.task_state == \
+                            'SHOULD BE KILLED':
+                        # This task has been stuck for longer than the
+                        # specified timeout. Attempt to kill it again.
+                        self.kill_task(task_id)
+                        continue
+
+                    if md.task_state not in (
                         'TASK_STAGING',
                     ):
                         continue
 
-                    md = self.task_metadata[task_id]
                     if time_now > (
                         md.task_state_history['TASK_STAGING'] +
                         self.task_staging_timeout_s
@@ -568,23 +593,40 @@ class ExecutionFramework(Scheduler):
 
         md = self.task_metadata[task_id]
 
-        # If we attempt to accept an offer that has been invalidated by master
-        # for some reason such as offer has been rescinded or we have exceeded
-        # offer_timeout, then we will get TASK_LOST status update back from
-        # mesos master.
-        if task_state == 'TASK_LOST' and \
-                'REASON_INVALID_OFFERS' == str(update.reason):
-            # This task has not been launched. Therefore, we are going to
-            # reenqueue it. We are not propogating any event up to the
-            # application.
-            log.warning('Received TASK_LOST from mesos master because we '
-                        'attempted to accept an invalid offer. Going to re-'
-                        'enqueue this task {id}'.format(id=task_id))
-            self.task_metadata = self.task_metadata.discard(task_id)
-            self.enqueue_task(md.task_config)
-            get_metric(TASK_LOST_DUE_TO_INVALID_OFFER_COUNT).count(1)
-            driver.acknowledgeStatusUpdate(update)
-            return
+        if task_state == 'TASK_LOST':
+            # If we attempt to accept an offer that has been invalidated by
+            # master for some reason such as offer has been rescinded or we
+            # have exceeded offer_timeout, then we will get TASK_LOST status
+            # update back from mesos master.
+            if 'REASON_INVALID_OFFERS' == str(update.reason):
+                # This task has not been launched. Therefore, we are going to
+                # reenqueue it. We are not propogating any event up to the
+                # application.
+                log.warning('Received TASK_LOST from mesos master because we '
+                            'attempted to accept an invalid offer. Going to '
+                            're-enqueue this task {id}'.format(id=task_id))
+                with self._lock:
+                    self.task_metadata = self.task_metadata.discard(task_id)
+                    self.enqueue_task(md.task_config)
+                get_metric(TASK_LOST_DUE_TO_INVALID_OFFER_COUNT).count(1)
+                driver.acknowledgeStatusUpdate(update)
+                return
+
+            elif str(update.message) == 'Reconciliation: Task is unknown':
+                # This could be a duplicate status update; we need to
+                # check if this is a duplicate update or not
+                if md.task_state == 'UNKNOWN':
+                    with self._lock:
+                        log.warning('Reenquing task {id} because we failed to '
+                                    'launch the task'.format(
+                                        id=task_id
+                                    ))
+                        self.task_metadata = self.task_metadata.discard(
+                            task_id
+                        )
+                        self.enqueue_task(md.task_config)
+                driver.acknowledgeStatusUpdate(update)
+                return
 
         self.event_queue.put(
             self.translator(update, task_id).set(task_config=md.task_config)
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -146,8 +146,7 @@ class ExecutionFramework(Scheduler):
                                     )
                                     )
                         # Re-enqueue task
-                        self.enqueue_task(self.task_metadata[task_id].
-                                          task_config)
+                        self.enqueue_task(md.task_config)
                         get_metric(TASK_LAUNCH_FAILED_COUNT).count(1)
                         continue
 
@@ -643,7 +642,7 @@ class ExecutionFramework(Scheduler):
                         'attempted to accept an invalid offer. Going to '
                         're-enqueue this task {id}'.format(id=task_id))
             # Re-enqueue task
-            self.enqueue_task(self.task_metadata[task_id].task_config)
+            self.enqueue_task(md.task_config)
             get_metric(TASK_LOST_DUE_TO_INVALID_OFFER_COUNT).count(1)
             driver.acknowledgeStatusUpdate(update)
             return
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -1,4 +1,5 @@
 import logging
+import socket
 import threading
 import time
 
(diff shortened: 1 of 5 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -115,31 +119,79 @@ def kill_tasks_stuck_in_staging(self):
...
+            for task_id in self.task_metadata.keys():
+                md = self.task_metadata[task_id]
+
+                eligible_for_termination_at = md.task_state_history[
+                    md.task_state] + self.task_staging_timeout_s
+
+                if time_now < eligible_for_termination_at:
+                    continue
+
+                if md.task_state == 'UNKNOWN':
```
huadongliu: TASK_UNKNOWN is used elsewhere.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -115,31 +119,79 @@ def kill_tasks_stuck_in_staging(self):
...
+
+        if len(tasks_to_reconcile) > 0:
+            try:
+                self.driver.reconcileTasks(tasks_to_reconcile)
+            except (socket.timeout, Exception):
+                log.warning('Failed to reconcile task status')
+
+        self._reconcile_tasks_at += time.time()
+
+    def _reenqueue_task(self, task_id):
```
huadongliu: enqueue_task will reinitialize task_metadata[task_id]. Is discard() needed?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -115,31 +119,79 @@ def kill_tasks_stuck_in_staging(self):
             if self.stopping:
```
huadongliu: Rename kill_tasks_stuck_in_staging?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -115,31 +119,79 @@ def kill_tasks_stuck_in_staging(self):
...
+                    self._reenqueue_task(task_id)
+                    get_metric(TASK_LAUNCH_FAILED_COUNT).count(1)
+                    continue
 
+                if md.task_state not in (
+                    'TASK_STAGING',
+                ):
+                    continue
+
+                if time_now > (
```
huadongliu: Line 144-149 can be combined as (task_state==STAGING && timed out) for easier read.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -115,31 +119,79 @@ def kill_tasks_stuck_in_staging(self):
...
+            tasks=tasks_to_reconcile
+        ))
+
+        if len(tasks_to_reconcile) > 0:
+            try:
+                self.driver.reconcileTasks(tasks_to_reconcile)
+            except (socket.timeout, Exception):
+                log.warning('Failed to reconcile task status')
+
+        self._reconcile_tasks_at += time.time()
```
huadongliu: The increment is task_reconciliation_delay, not time.time().

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -115,31 +119,79 @@ def kill_tasks_stuck_in_staging(self):
...
-                            'Killing stuck task {id}'.format(id=task_id)
-                        )
-                        self.kill_task(task_id)
-                        self.blacklist_slave(
-                            agent_id=self.task_metadata[task_id].agent_id,
-                            timeout=self.slave_blacklist_timeout_s,
-                        )
-                        get_metric(TASK_STUCK_COUNT).count(1)
+            time_now = time.time()
+            for task_id in self.task_metadata.keys():
```
huadongliu: Iterating task_metadata needs to be locked since new tasks may come.

task_processing/plugins/mesos/execution_framework.py
sagar8192: No, I have used UNKNOWN everywhere.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Yeah, makes sense. Changed it.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Good catch!

task_processing/plugins/mesos/execution_framework.py
sagar8192: Good catch! Fixed it.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -568,29 +630,26 @@ def statusUpdate(self, driver, update):
...
             # application.
             log.warning('Received TASK_LOST from mesos master because we '
-                        'attempted to accept an invalid offer. Going to re-'
-                        'enqueue this task {id}'.format(id=task_id))
-            self.task_metadata = self.task_metadata.discard(task_id)
-            self.enqueue_task(md.task_config)
+                        'attempted to accept an invalid offer. Going to '
+                        're-enqueue this task {id}'.format(id=task_id))
+            # Re-enqueue task
+            self.enqueue_task(self.task_metadata[task_id].task_config)
```
huadongliu: md is self.task_metadata[task_id].

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -527,20 +572,37 @@ def resourceOffers(self, driver, offers):
...
+                            'Thus, moving them to UNKNOWN state'.format(
+                                tasks=', '.join([
+                                    task.task_id.value for task in
+                                    tasks_to_launch
+                                ]),
+                            )
+                            )
+                task_launch_failed = True
+
+            current_task_state = 'UNKNOWN' if task_launch_failed else \
```
huadongliu: mesos_status_to_event uses TASK_UNKNOWN to map mesos state to an event.

sagar8192: This is different than TASK_UNKNOWN. We set the state of tasks to 'UNKNOWN' that are not launched successfully. When we reconcile the state of these tasks, mesos will return TASK_LOST.

huadongliu: You are right. This UNKNOWN is not put to executor's queue. It would be helpful to add a comment on this.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -96,7 +102,7 @@ def __init__(
 
         self._initialize_metrics()
         self._last_offer_time = None
-        self._task_states = {
+        self._terminal_task_states = {
```
Rob-Johnson: terminal_task_counts?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -106,28 +112,45 @@ def __init__(
...
+                                   self.task_staging_timeout_s):
+                        continue
+
+                    if md.task_state == 'UNKNOWN':
+                        log.warning('Task {id} has been in unknown state for '
+                                    'longer than {timeout}. Re-enqueuing it.'
+                                    .format(
+                                        id=task_id,
+                                        timeout=self.task_staging_timeout_s
+                                    )
```
Rob-Johnson: minor: this indentation seems off. is this what autopep8 gave you?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -106,28 +112,45 @@ def __init__(
...
 
+            time_now = time.time()
             with self._lock:
-                time_now = time.time()
                 for task_id in self.task_metadata.keys():
-                    if self.task_metadata[task_id].task_state not in (
+                    md = self.task_metadata[task_id]
+
+                    if md.task_state not in (
                         'TASK_STAGING',
```
Rob-Johnson: this isn't a pattern introduced by you, but these magic strings kind of bug me. We should define constants for these things.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -138,8 +161,28 @@ def kill_tasks_stuck_in_staging(self):
                         )
                         get_metric(TASK_STUCK_COUNT).count(1)
 
+            self._reconcile_tasks()
             time.sleep(10)
 
+    def _reconcile_tasks(self):
```
Rob-Johnson: can you parameterise {{tasks_to_reconcile}} so that we can also use this to run periodic *implicit* reconciliations, as well as these more deliberate explicit reconciliations?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -138,8 +161,28 @@ def kill_tasks_stuck_in_staging(self):
...
+        if time.time() < self._reconcile_tasks_at:
```
Rob-Johnson: not a blocker, but long term I'd like the decision of 'when' to reconcile to be the responsibility of the caller, so that we can have an implicit reconciliation on both a regular period basis (every 15 mins) as well as the explicit reconciliation required when we're asking mesos about specific tasks.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -138,8 +161,28 @@ def kill_tasks_stuck_in_staging(self):
...
+            return
+
+        tasks_to_reconcile = [Dict({'task_id': task_id}) for
+                              task_id in self.task_metadata]
+
+        log.info('Reconciling following tasks {tasks}'.format(
+            tasks=tasks_to_reconcile
+        ))
+
+        if len(tasks_to_reconcile) > 0:
```
Rob-Johnson: I don't think you need this - it's fine to reconcile with an empty list (and ask mesos for a status update on *every* task)

task_processing/runners/subscription.py
```diff
@@ -33,7 +33,7 @@ def run(self, task_config):
         return self.executor.run(task_config)
 
     def kill(self, task_id):
-        return self.executor.kill(task_id)
+        self.executor.kill(task_id)
```
Rob-Johnson: why don't we want to return here?

task_processing/plugins/mesos/execution_framework.py
sagar8192: Sure.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Sounds good. The current setup is for implicit invocation. We can change this if there is a need for explicit invocation.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Pymesos is not handling it correctly, that's why we have to do it. https://github.com/douban/pymesos/blob/cd6569cc5d167a8edf819c569a2c61dd05e4053e/pymesos/scheduler.py#L316

task_processing/plugins/mesos/execution_framework.py
sagar8192: Yeah, I agree. Created a separate ticket: TASKPROC-122

task_processing/runners/subscription.py
sagar8192: kill is not synchronous, so it should not return anything.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Sure.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -527,20 +568,39 @@ def resourceOffers(self, driver, offers):
...
-                            task_state='TASK_STAGING',
+                            task_state=current_task_state,
                             task_state_history=md.task_state_history.set(
-                                'TASK_STAGING', time.time()),
+                                current_task_state, time.time()),
+
                         )
                     )
-                    get_metric(TASK_LAUNCHED_COUNT).count(1)
+                    if not task_launch_failed:
```
keymone: we should probably be counting number of failures as well, sounds like very useful statistic

task_processing/runners/subscription.py
keymone: i prefer everything returning and everything having a value. less things to fix when implementation changes.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Good call. Added a metric for this.

PR #75, opened 2017-10-02
Catch stateful_executor exceptions to keep events flowing

huadongliu:
Tested with examples/dynamo_persistence.py with a wrong table name.

### DIFF
--- a/task_processing/plugins/stateful/stateful_executor.py
+++ b/task_processing/plugins/stateful/stateful_executor.py
@@ -1,8 +1,12 @@
+import logging
 import threading
+import traceback
 
 from six.moves.queue import Queue
 
 from task_processing.interfaces.task_executor import TaskExecutor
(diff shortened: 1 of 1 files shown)


task_processing/plugins/stateful/stateful_executor.py
```diff
@@ -41,6 +45,10 @@ def get_event_queue(self):
     def subscribe_to_updates_for_task(self):
```
sagar8192: Can we wrap the whole function to prevent the thread from crashing?

huadongliu: Done.

sagar8192: Dicussed with @huadongliu in person and it seems better to just catch Dynamodb exceptions.

PR #77, opened 2017-10-05
Fixed the exception when an agent being blacklisted more than once

huadongliu:
This fixes TASKPROC-129. 

Another way to fix it is to merge blacklisting with the background thread. I am keeping the original design to make the processing simple at the cost of one sleeping thread for each blacklisted agent.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -194,11 +194,8 @@ class ExecutionFramework(Scheduler):
 
     def blacklist_slave(self, agent_id, timeout):
         with self._lock:
-            if agent_id in self.blacklisted_slaves:
-                # Punish this slave for more time.
-                self.blacklisted_slaves = \
-                    self.blacklisted_slaves.remove(agent_id)
-
+            # A new entry is appended even if the agent is being blacklisted.
+            # This is equivalent to restarting the blacklist timer.
             log.info('Blacklisting slave: {id} for {secs} seconds.'.format(
                 id=agent_id,
                 secs=timeout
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
I think this is ok, but I'm not sure how this is happening if we wrap the call with ``with self._lock`` and we also wrap the body of ``unblacklist_slave`` with the same lock ? Can you explain more about the root cause?

keymone:
@robj i think it's because we make a new thread every time we blacklist an agent, so even though we drop redundant agent_id from the list we never kill the thread that is waiting to call another .remove.

imo i'd just reimplement that bit to run in a single thread that iterates over blacklisted agents and removes those that are expired.

Rob-Johnson:
ah yeah, you're right @keymone - see the stacktrace

```
Exception in thread Thread-5:
--
Traceback (most recent call last):
File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
self.run()
File "/usr/lib/python3.6/threading.py", line 864, in run
self._target(*self._args, **self._kwargs)
File "/nail/home/kdeal/repos/jolt/venv/lib/python3.6/site-packages/task_processing/plugins/mesos/execution_framework.py", line 182, in unblacklist_slave
ValueError: PVector.remove(x): x not in vector
```

its in unblacklist that this fails.

so couldn't we just wrap

https://github.com/Yelp/task_processing/blob/65b3c60a35ba67d89c1bf7ef273a6b6b3f3aaa9b/task_processing/plugins/mesos/execution_framework.py#L218-L219

with 
```
if agent_id in self.blacklisted_slaves:
    self.blacklisted_slaves = \
```

Rob-Johnson:
as for the single thread vs multi-thread thing - I kind of like that the thread per slave thing gives us more accurate removal (timer per thread, rather than a single thread checking at some interval)

keymone:
we could wrap in a check but this PR is ok as well, i'll give it a ship.

huadongliu:
@Rob-Johnson Yes, checking before removing will work. I didn't do it because the agent is just penalized once when it should be penalized longer.

sagar8192:
+1 for a single thread that iterates over the map of blacklisted slaves and un-blacklists the slaves. This can be a part of background thread that we already have. I am not a fan of having multiple threads.

PR #78, opened 2017-10-12
Decline offers in batch, fix reconciliation with 300s interval

huadongliu:
Tests on norcal-devc show /master/api/v1/scheduler hits are reduced a lot.


huadongliu:
This PR aims to reduce the scheduler endpoint hits. In addition to combining offer declines, I felt reconciliation for every minute is a bit too frequent. I picked 5 minutes, but I am fine with other numbers, say 3 minutes.

keymone:
@huadongliu 5 minutes is probably ok, ship!

PR #80, opened 2017-10-20
Add retrying executor example and pass params on task_config

chlgit:
Add retrying executor at examples/retry.py

Modify retrying executor to allow user passing params through task_config like this:
task_config = TaskConfig(
        image='docker-dev.yelpcorp.com/dumb-busybox',
        cmd='/bin/false',
        retries=5,
        retry_pred='lambda e: not e.success'
    )

Manually testing by command 
"docker-compose -f examples/cluster/docker-compose.yaml run playground examples/retry.py"

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -68,6 +68,10 @@ class MesosTaskConfig(PRecord):
                     factory=float,
                     mandatory=False,
                     invariant=lambda t: (t > 0, 'timeout > 0'))
+    retries = field(type=int,
+                    factory=int,
+                    mandatory=False,
+                    invariant=lambda r: (r > 0, 'retries >= 0'))
     volumes = field(type=PVector,
                     initial=v(),
                     factory=pvector,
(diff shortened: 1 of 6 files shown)


chlgit:
Discussed with @sagar8192. We would like to leave the extension field for another ticket if we really decide to move to that way. It looks like only timeout and retry should move to extension field at this point, and code modifications for extensions field seems to be out of the scope of this ticket. 
@keymone any thought?

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -68,6 +68,10 @@ def __invariant__(conf):
                     factory=float,
                     mandatory=False,
                     invariant=lambda t: (t > 0, 'timeout > 0'))
+    retries = field(type=int,
+                    factory=int,
+                    mandatory=False,
+                    invariant=lambda r: (r > 0, 'retries > 0'))
```
keymone: **re**tries reads as "number of tries after failure", constraining this to `> 0` means users have no way to say "don't retry this even once". change constrain to `>=`?

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -41,25 +41,27 @@ def event_with_retries(self, event):
         )
 
     def retry(self, event):
-        current_retry_attempt = self.task_retries[event.task_id]
+        remain_retry_attempt = self.task_retries[event.task_id]
```
keymone: rename to `retries_remaining`

PR #81, opened 2017-10-26
Fixed retry logging and retries counts

huadongliu:
1. Fix incorrect retrying log message
2. Make the field "retries" truly the number of retries, not the total number of runs (retries+1).

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -68,10 +68,12 @@ class MesosTaskConfig(PRecord):
                     factory=float,
                     mandatory=False,
                     invariant=lambda t: (t > 0, 'timeout > 0'))
+    # By default, the retrying executor retries 3 times. This task option
+    # overrides the executor setting.
     retries = field(type=int,
                     factory=int,
                     mandatory=False,
-                    invariant=lambda r: (r > 0, 'retries >= 0'))
+                    invariant=lambda r: (r >= 0, 'retries >= 0'))
     volumes = field(type=PVector,
                     initial=v(),
                     factory=pvector,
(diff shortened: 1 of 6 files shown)


huadongliu:
We already have invariant of retries as "retries > 0". Did I miss anything? 0https://github.com/Yelp/task_processing/blob/master/task_processing/plugins/mesos/mesos_executor.py#L71

task_processing/plugins/mesos/retrying_executor.py
```diff
@@ -102,9 +106,9 @@ def retry_loop(self):
     def run(self, task_config):
         if task_config.task_id not in self.task_retries:
             with self.task_retries_lock:
-                retry_value = self.retries - 1
+                retry_value = self.retries
                 if 'retries' in task_config:
```
keymone: this construct is a bit confusing, let's go with simple if, same as on line 48: https://github.com/Yelp/task_processing/pull/81/files#diff-b78ffcc2ae6bff2d45f74a6fe10dc8ebR48

in fact just move it into separate function

    def task_or_executor_retries(self, task_config):
        return task_config.retries if 'retries' in task_config else self.retries

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -71,7 +71,7 @@ def __invariant__(conf):
     retries = field(type=int,
                     factory=int,
                     mandatory=False,
-                    invariant=lambda r: (r > 0, 'retries >= 0'))
+                    invariant=lambda r: (r > 0, 'retries > 0'))
```
keymone: this is still wrong in my opinion, see https://github.com/Yelp/task_processing/pull/80#discussion_r146798275

tl;dr since retries is not including the original run and the lowest possible value user can provide is `1`, user has no way of saying "i don't want this to be retried". invariant predicate should be changed to `r >= 0`.

PR #82, opened 2017-10-30
Added an option to capture remote task logs

huadongliu:
This is TASKPROC-127. 

- I wish I could use mesos-cli to hit the mesos files/read endpoint. However, this would create a circular dependency between paasta and taskproc.
- I chose to fetch task log in the framework scheduler instead of in an executor plugin because I feel logging is part of the remote execution.
- I chose to use the single background thread to fetch logs periodically instead of downloading the whole log at the end to avoid a one time spike on the mesos API. One thread per task is not used because it will mess up output.

### DIFF
--- a/examples/sync.py
+++ b/examples/sync.py
@@ -27,6 +27,12 @@ def main():
 
     TaskConfig = executor.TASK_CONFIG_INTERFACE
     task_config = TaskConfig(image="busybox", cmd='/bin/true')
+    # Test log fetching
+    # task_config = TaskConfig(
+    #     image="ubuntu:14.04",
+    #     task_log=True,
+    #     cmd="bash -c 'for i in $(seq 1 30); do echo $i&&sleep 10; done'"
+    # )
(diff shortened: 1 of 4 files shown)


keymone:
i think this should be a separate executor. i know it will be an executor that can't be reused for other backends, but whenever there is some isolated piece of functionality that we can move out of the main framework - we should move it out of the main framework.

i also think task_log attribute on task_config is non-essential, i'd really like us to start moving things into extensions attribute instead of widening the interface with every pull request.

huadongliu:
Unfortunately, log fetching has to sit at the very bottom of the stack because it needs to know details at the task level. It won't work with a separate executor X. For example, if the retrying executor being the downstream executor of X, X won't be able to fetch logs of the retries because it has no knowledge of those retries. You can possibly make it work. However, it will violate layering rules.

huadongliu:
Closing this PR in favor of an implementation in an executor.

PR #83, opened 2017-10-31
Added an executor to stream task logs

huadongliu:


task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,151 @@
...
+        # No lock is needed because log_metadata is touched in a single thread
+        self.log_metadata = m()
+
+        self.logging_thread = Thread(target=self.logging_loop)
+        self.logging_thread.daemon = True
+        self.logging_thread.start()
+
+    def build_task_log_url(self, task_log_url, event):
+        try:
+            response = json.load(urllib.request.urlopen(
```
Rob-Johnson: minor nit: can we use requests instead of urllib?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,151 @@
...
+        self.log_metadata = self.log_metadata.set(
+            task_id,
+            log_md.set(log_offsets=pmap(offsets)),
+        )
+
+    def logging_loop(self):
+        while True:
+            # process downstream events
+            while not self.src_queue.empty():
+                e = self.src_queue.get()
```
Rob-Johnson: do you need to call ``src_queue.task_done`` when you're finished?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,151 @@
...
+                self.dest_queue.put(e)
+                # Record the 1st part of the log url
+                if e.kind == 'task' and e.platform_type == 'staging':
```
Rob-Johnson: do you need to have separate processing at staging vs running? couldn't you do all of the url creation when you get the first TASK_RUNNING, and ignore any further events you get for the same task id?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,151 @@
...
+            while True:
+                url = log_md.log_url + f + \
+                    '&length=' + str(TASK_LOG_CHUNK_LEN) + \
+                    '&offset=' + str(offset)
+                try:
+                    response = json.load(urllib.request.urlopen(url))
+                    log_length = len(response['data'])
+                    for line in response['data'].splitlines():
+                        print(
+                            task_id + ": " + line,
```
Rob-Johnson: I think we're going to want the log line to be as 'rich' as possible.

Can we do things like format it as

iso8601datetime task_id[container_id@agent_ip]: logline

task_processing/plugins/mesos/__init__.py
```diff
@@ -8,6 +9,7 @@
 
 def register_plugin(registry):
     return registry \
+        .register_task_executor('logging', LoggingExecutor) \
```
Rob-Johnson: Can we call it the 'MesosLoggingExecutor' to try reset expectations about what this really is?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,151 @@
...
+                    )
+
+                # Fetch the last log and remove the entry
+                if e.kind == 'task' and e.terminal:
+                    self.stream_task_log(e.task_id)
+                    self.log_metadata.discard(e.task_id)
+
+            # grab logs
+            for task_id in self.log_metadata.keys():
+                self.stream_task_log(task_id)
```
keymone: i don't think we can afford to fetch logs in this thread as we'd be blocking the queue chain

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -609,7 +609,19 @@ def resourceOffers(self, driver, offers):
 
                         )
                     )
+                    # Emit the staging event for successful launches
                     if not task_launch_failed:
+                        update = Dict(
```
keymone: i know why we need this, but it feels too arbitrary. we could instead attach task offer to each status update (or at least to task_running) in the framework - this way status updates are all uniform and every executor has the same data to work with.

task_processing/plugins/mesos/logging_executor.py
huadongliu: Good catch! I will fix it along with other comments from you and @Rob-Johnson.

task_processing/plugins/mesos/__init__.py
huadongliu: Will fix.

task_processing/plugins/mesos/execution_framework.py
huadongliu: Will fix.

task_processing/plugins/mesos/logging_executor.py
huadongliu: Will fix.

task_processing/plugins/mesos/logging_executor.py
huadongliu: Will fix except for the iso time. Applications should add timestamps to their logs instead. The executor timestamp is not accurate.

task_processing/plugins/mesos/logging_executor.py
huadongliu: The staging event passes up the agent address (offer). This info will not repeat in the running event.

task_processing/plugins/mesos/logging_executor.py
huadongliu: It is not required since we do not use join() for any kind of synchronization. I can still add it though.

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,208 @@
...
+        offsets = {
+            'stdout': log_md.log_offsets['stdout'],
+            'stderr': log_md.log_offsets['stderr']
+        }
+        agent = urlparse(log_md.log_url).hostname
+
+        for f in ['stdout', 'stderr']:
+            offset = offsets[f]
+            log_path = log_md.log_path + "/" + f
+            while True:
```
keymone: it could happen that stderr will only be printed after stdout and users will have to order messages, which is not great. but we can fix it later when it actually becomes problem people complain about.

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,208 @@
...
+                task_id = self.done_tasks[0]
+                self.stream_task_log(task_id)
+                with self.task_lock:
+                    self.done_tasks = self.done_tasks.remove(task_id)
+                    self.running_tasks = self.running_tasks.discard(task_id)
+
+            if self.stopping:
+                return
+
+            time.sleep(10)
```
keymone: maybe reduce to sleep(1)? is files/ endpoint generating too much load on mesos?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -0,0 +1,208 @@
...
+
+        self.event_thread = Thread(target=self.event_loop)
+        self.event_thread.daemon = True
+        self.event_thread.start()
+
+        self.logging_thread = Thread(target=self.logging_loop)
+        self.logging_thread.daemon = True
+        self.logging_thread.start()
+
+    def get_task_log_path(self, task_id):
```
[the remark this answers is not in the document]
huadongliu: will fix.

task_processing/plugins/mesos/logging_executor.py
huadongliu: You are right. I wanted to reduce the number of file API hits as much as possible. Mesos API is single threaded. I don't know exactly how expensive the file API is, but we have seen a significant impact of state.json (the difference is just serialization?).

keymone: ok, let's run with 10s for now and iterate.

PR #86, opened 2017-11-10
Make queues in taskproc framework unbounded.

sagar8192:


Rob-Johnson:
@sagar8192 can you point us to an issue or something to justifies this? on the face of it, unbounded queues don't seem sensible, but I guess there was a specific problem you had to solve?

keymone:
cc @kkellyy

sagar8192:
@Rob-Johnson @keymone Sorry, forgot to link the ticket. Here it is: RELENG-24551. We were seeing an issue in jolt where they were running ~2.5k bundles and the driver thread was getting stuck. We investigated the issue and found that the driver thread was getting stuck while adding data to the status update queue. We removed the limits to prevent that from happening. Jolt is now able to successfully run the bundles after the fix. It would be better to store the data in queue rather than blocking the driver thread. Also, other queues in our executor stack are not bounded, so it should be okay to have unbounded queues in the framework.

Rob-Johnson:
hm, I'm not totally convinced that making the queues unbounded is the long-term answer.

Should we figure out a strategy for implementing back pressure if the queue of tasks to launch gets too big? We're probably fine so long as we're getting offers that can place the tasks, but I'm concerned about what'll happen if a user queues up an unbounded number of tasks to be processed. Is the 2.5k bundles a 'normal' number for Jolt to launch?

anyway, I guess this is fine for now - lets figure out a longer term strategy though.

ssk2:
cc @solarkennedy

PR #87, opened 2017-12-21
More verbose logging of offers rejected due to pool mismatch

keymone:


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -188,10 +188,11 @@ def offer_matches_pool(self, offer):
         if self.pool is None:
             # If pool is not specified, then we can accept offer from any agent
             return True
```
huadongliu: This will break Line 561.

PR #88, opened 2018-01-26
Extend kill_task to remove queued tasks

sagar8192:
Manually tested

Added a unit test

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -196,7 +196,22 @@ class ExecutionFramework(Scheduler):
         return False, None
 
     def kill_task(self, task_id):
-        self.driver.killTask(Dict(value=task_id))
+        tmp_list = []
+        flag = False
+        with self._lock:
+            while not self.task_queue.empty():
+                t = self.task_queue.get()
+                if task_id == t.task_id:
+                    flag = True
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -196,7 +196,22 @@ def offer_matches_pool(self, offer):
...
     def kill_task(self, task_id):
-        self.driver.killTask(Dict(value=task_id))
+        tmp_list = []
+        flag = False
+        with self._lock:
+            while not self.task_queue.empty():
+                t = self.task_queue.get()
+                if task_id == t.task_id:
+                    flag = True
+                    self.task_metadata = self.task_metadata.discard(task_id)
```
huadongliu: Why don't you break from the loop when a match is found? Do you want to keep the order of tasks? I am fine with this implementation. You may want to consider an implementation that does actions according to task states, for example, no-op if the task is not in task_metadata, loop queue to delete if the task is in TASK_INITED state, and call driver's kill otherwise.

sagar8192: Yes, I want to preserve the order of tasks. Yeah, I considered that option. But, it requires more work. The current implementation is simpler, IMO.

sagar8192:
FYI, @kkellyy

PR #90, opened 2018-02-26
[WIP] Support mesos pods

sagar8192:
This is not complete yet, but I would like to get your feedback about the approach/design. 

This PR add support for mesos pods to taskproc. I have made following changes: 
1) Changed the datastructures in execution framework to have extra data for pods. 
2) Changed the status update to have task_id and status update mappings. Mesos sends status updates for individual tasks in pod, so I have added logic to combine these and send an event when either all the tasks have reached the same state or a terminal state.

What is outstanding:
1) Fix/add unit tests
2) Need to change the kill API.
3) Make changes to retrying, timeout and persistent executor.


huadongliu:
I am wondering whether we have to explicitly call out MesosPodConfig. Can we assume a single task a pod that only has one task?  Then we can have a single code flow to enqueue and launch.

An alternative is to infer pod or task by looking at len(tasks) and then construct launch group or individual task info separately. It is similar to this PR without calling out MesosPodConfig explicitly.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+            for task_id in self._pod_to_task_mappings[pod_id]:
+                tm.set(
+                    task_id,
+                    self.task_metadata[task_id].task_config
+                )
+
+        return tm
+
+    def _update_internal_state(self, task_id, task_state, metadata, update):
+        is_terminal = True if task_state in self._terminal_task_counts \
```
Rob-Johnson: just 

```
is_terminal = task_state in self._terminal_task_counts
```

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+                # Check if all the tasks in the pod have received a
+                # terminal status update or not.
+                if md.terminated is False:
+                    all_tasks_terminated = False
+
+                if task_state not in md.task_state_history:
+                    state_changed = False
+
+            if all_tasks_terminated is True or state_changed is True:
+                # This means that all the nested containers in the pod have
```
Rob-Johnson: the comment doesn't align with what is actually happening? you're doing this on a state change for *any* task in the pod afaict?

task_processing/interfaces/event.py
```diff
@@ -9,10 +9,12 @@
 from pyrsistent import PRecord
 
 
-EVENT_KINDS = ['task', 'control']
+EVENT_KINDS = ['task', 'control', 'pod']
 
 
 class Event(PRecord):
+    # TODO(sagarp): Add an invariant for task_id_task_config_mappings and
```
Rob-Johnson: please either do this or remove the TODO

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -48,6 +50,9 @@
 
 class TaskMetadata(PRecord):
     agent_id = field(type=str, initial='')
+    pod_id = field(type=(str, type(None)), initial=None)
+    terminated = field(type=bool, initial=False)
+    terminal_raw_status_update = field(type=(type(None), Dict), initial=None)
```
Rob-Johnson: what's this for? I can't see it being read anywhere, only written to?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+                    task_id,
+                    self.task_metadata[task_id].task_state_history
+                    [task_state]['status_update']
+                )
+
+        return sm
+
+    def _aggregate_task_configs_for_pod(self, pod_id):
+        tm = m()
+        with self._lock:
```
Rob-Johnson: why do you need to get the lock here?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+            task_infos = task_operation.launch.task_infos
+        if len(task_group_operations) > 0:
+            for task in task_group_operations:
+                task_infos = task_infos + task.launch_group.task_group.tasks
+
+        return task_infos
+
+    def _aggregate_status_updates_for_pod(self, pod_id, task_state):
+        sm = m()
+        with self._lock:
```
Rob-Johnson: why do you need the lock here?

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -74,6 +74,8 @@ def __invariant__(conf):
                     factory=int,
                     mandatory=False,
                     invariant=lambda r: (r >= 0, 'retries >= 0'))
+    cni_network = field(type=str, initial='cni-test')
```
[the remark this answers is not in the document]
Rob-Johnson: can you also not set the initial here? otherwise every task is going to have this attached to it.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -239,28 +245,76 @@ def unblacklist_slave(self, agent_id, timeout):
...
+                                TASK_INITED=m(
+                                    time_stamp=time.time(),
+                                    status_update=Dict()
+                                )
+                            ),
+                        )
+                    )
+                    nested_container_ids.append(task.task_id)
+
+                self._pod_to_task_mappings = self._pod_to_task_mappings.set(
```
Rob-Johnson: do we really need another datastructure for this? I know it makes for an easier lookup in task_metadata, but i'm kind of wondering if it'd be cleaner to just have a function, since in the end these are always used to access task_metadata? it reduces the two sources of truth to one:

```
def task_for_pod_ids(pod_id):
    return [tm for tm in self.task_metadata.items() if tm.value.pod_id == pod_id]
```

task_processing/interfaces/event.py
```diff
@@ -35,6 +37,13 @@ class Event(PRecord):
         invariant=lambda x: (isinstance(x, PMap),
                              'task_config must inherit from PMap'),
         factory=lambda x: pmap(x) if not isinstance(x, PMap) else x)
+
+    # This fields are only applicable to event of type pod
+    task_config_per_task_id = field(type=PMap)
```
Rob-Johnson: I don't really like this. The whole point in the interface is that its platform generic, and yet you're adding more keys here. Why do you need to return all the task_configs for tasks in the pod if you're issuing a pod specific event? same with ``mesos_status_per_task_id``

task_processing/plugins/mesos/execution_framework.py
sagar8192: Good catch! Removed it.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
             container=container
         )
 
+        if task_config.health_check is True:
+            task_info.health_check = Dict(
```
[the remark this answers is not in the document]
sagar8192: I changed the name from health_check to http_healthcheck_port. HTTP and TCP health checks for nested containers are not going to work after my change to mesos. The TCP and HTTP health checks for pods are implemented with the assumption that health check process runs in the same network namespace as the target container. The command check, however, runs in the same namespace as the target nested container. So, if people want to use HTTP health check, then they can specify the port number. We can make the command customizable in future if anyone requests it.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Yeah, we can certainly do it this way. But, HashMap is more efficient than doing this. It is just one extra data structure.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Removed it.

task_processing/plugins/mesos/execution_framework.py
sagar8192: Removed it.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -239,28 +245,76 @@ def unblacklist_slave(self, agent_id, timeout):
             self.blacklisted_slaves = \
                 self.blacklisted_slaves.remove(agent_id)
 
-    def enqueue_task(self, task_config):
+    def _is_pod(self, task_config):
```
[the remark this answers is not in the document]
sagar8192: Good call.

task_processing/plugins/mesos/execution_framework.py
sagar8192: ++

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+                )
+
+            if all_tasks_terminated is True:
+                # Discard the pod since it has reached a terminal state
+                with self._lock:
+                    for t in self._pod_to_task_mappings[metadata.pod_id]:
+                        self.task_metadata = \
+                            self.task_metadata.discard(t)
+
+                    self._pod_to_task_mappings.discard(metadata.pod_id)
```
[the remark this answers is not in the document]
sagar8192: Good catch!

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+                    md = self.task_metadata[t]
+                # Check if all the tasks in the pod have received a
+                # terminal status update or not.
+                if md.terminated is False:
+                    all_tasks_terminated = False
+
+                if task_state not in md.task_state_history:
+                    state_changed = False
+
+            if all_tasks_terminated is True or state_changed is True:
```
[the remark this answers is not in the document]
sagar8192: Yes, we can just do this way as well. I find it easier to read if the assertion is explicitly checking for what is expected.

For eg, all_tasks_terminated can be equal to 1 and the above check will pass. So, checking if the value is True or not is better for both readability and correctness.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -441,6 +596,126 @@ def create_new_docker_task(self, offer, task_config, available_ports):
...
+                    [task_state]['status_update']
+                )
+
+        return sm
+
+    def _aggregate_task_configs_for_pod(self, pod_id):
+        tm = m()
+        with self._lock:
+            for task_id in self._pod_to_task_mappings[pod_id]:
+                tm.set(
```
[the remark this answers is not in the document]
sagar8192: Good catch!

sagar8192:
@keymone Our current stack is very task-centric. Adding support for pods means we have to change all the layers. For eg, the persistent executor needs to have a different schema to store the status of pods, the retrying executor needs to be able to distinguish between pods and tasks and so on. Putting the pod specific logic in a separate executor is also a good idea, but we still need some logic in the execution framework, which is necessary to build protobufs etc. So, we can't move all the logic into the pod executor. Also, people should be able to run both pods and tasks using the same stack, so putting the logic for pods at one place makes more sense. 
 
Pods behave differently than tasks. If a container in a pod exits with exit code than 0, then the entire pod gets killed. But, if a task in a pod exits with a status code 0, then the rest of containers continue to run. This is not ideal for the yelp-main use case. We will kill the entire the pod if any container exits.

Treating everything as a pod is interesting but not very practical. All of task_processing's clients will have to change their code to acommodate this.

PR #91, opened 2018-04-09
Use default 240 sec staging timeout for Mesos

vkhromov:


Rob-Johnson:
can you also add your fix to the dockerfile that upgraded setuptools? that should fix the tests, too

vkhromov:
@Rob-Johnson, please see #94 for test fixes and #95 for timeout changes.  Closing this one.

PR #95, opened 2018-04-09
Mesos 240 sec staging timeout

vkhromov:


task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -26,7 +27,7 @@ def __init__(
         initial_decline_delay=1.0,
         framework_translator=mesos_status_to_event,
         framework_name='taskproc-default',
-        framework_staging_timeout=60,
+        framework_staging_timeout=240,
```
Rob-Johnson: lets get rid of the default in ``execution_framework.py`` too

PR #97, opened 2018-04-13
Added EQUALS/NOTEQUALS/LIKE/UNLIKE constraints

vkhromov:


task_processing/plugins/mesos/constraints.py
```diff
@@ -0,0 +1,89 @@
+import abc
+import re
+
+from pyrsistent import field
+from pyrsistent import PRecord
+
+
+class Operator(object):
```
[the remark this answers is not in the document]
Rob-Johnson: +1 for using functions

task_processing/plugins/mesos/task_config.py
```diff
@@ -9,6 +9,8 @@
 from pyrsistent import pvector
 from pyrsistent import v
 
+from .constraints import valid_constraint_operator_name
```
Rob-Johnson: remove the .

task_processing/plugins/mesos/constraints.py
```diff
@@ -0,0 +1,89 @@
...
+    # the constraint is satisfied.
+    if actual_value is None:
+        return True
+
+    # The operator names have already been validated by the validator in
+    # `MesosTaskConfig`, so it's guaranteed that it's in `OPERATORS`.
+    return OPERATORS[c.operator](c.value, actual_value)
+
+
+def dict_matches_constraints(kv, constraints):
```
Rob-Johnson: can you call this 'attributes_match_all_constraints'?

task_processing/plugins/mesos/constraints.py
```diff
@@ -0,0 +1,89 @@
...
+    'EQUALS': EqualsOperator(),
+    '==': EqualsOperator(),
+    'NOTEQUALS': NotEqualsOperator(),
+    '!=': NotEqualsOperator(),
+    'LIKE': LikeOperator(),
+    'UNLIKE': UnlikeOperator(),
+}
+
+
+def _dict_matches_constraint(kv, constraint):
```
Rob-Johnson: call this 'attributes_match_constraint'

task_processing/plugins/mesos/task_config.py
```diff
@@ -83,6 +103,12 @@ def __invariant__(conf):
         mandatory=False,
         invariant=lambda t: (t > 0, 'timeout > 0')
     )
+    constraints = field(
+        type=PVector,
+        initial=v(),
+        factory=pvector,
```
Rob-Johnson: set this to be ```lambda v: Constraint(attribute=v[0], operator=v[1], value=v[2])```

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -14,6 +14,7 @@
 from pyrsistent import v
 from six.moves.queue import Queue
 
+from .constraints import dict_matches_constraints
```
Rob-Johnson: remove the . - you don't need a relative import

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -223,6 +224,11 @@ def offer_matches_pool(self, offer):
 
         return False, None
 
+    def offer_matches_task_constraints(self, offer, task):
```
Rob-Johnson: you could move this to ``constraints.py`` - it doesn't need to be part of this object

task_processing/plugins/mesos/constraints.py
solarkennedy: yes, and can you rename kv => attributes?

task_processing/plugins/mesos/constraints.py
```diff
@@ -0,0 +1,75 @@
...
+    return expected_value == actual_value
+
+
+def notequals_op(expected_value, actual_value):
+    return expected_value != actual_value
+
+
+def like_op(like_pattern, actual_value):
+    escaped_pattern = re.escape(like_pattern)
+    re_pattern = escaped_pattern.replace('\\%', '.*').replace('_', '.')
```
Rob-Johnson: what is this line for?

task_processing/plugins/mesos/constraints.py
```diff
@@ -0,0 +1,75 @@
...
+def attributes_match_constraints(attributes, constraints):
+    # If constraints aren't specified then they are satisfied.
+    if constraints is None:
+        return True
+
+    return all(_attributes_match_constraint(attributes, c)
+               for c in constraints)
+
+
+_OPERATORS_NAMES = OPERATORS.keys()
```
Rob-Johnson: you probably don't need this if you're only using it once?

vkhromov:
@keymone, @Rob-Johnson, @sagar8192, @solarkennedy: please see the updated diff which addresses your review comments.

task_processing/plugins/mesos/task_config.py
```diff
@@ -23,6 +27,17 @@ def valid_volumes(volumes):
     return (True, None)
 
 
+def _valid_constraints(constraints):
+    for constraint in constraints:
+        operator_name = constraint.operator
+        if not valid_constraint_operator_name(operator_name):
+            return (
```
Rob-Johnson: rather than exiting on the first failure, can you iterate over all the constraints to see which would have been invalid? That way the user gets to see all the failures up front.

task_processing/plugins/mesos/constraints.py
keymone: that's the wildcard syntax for SQL's LIKE operator

vkhromov: Yeah, eventually we agreed that it's probably better to use old good regexs instead of SQL LIKE here. :)

PR #99, opened 2018-04-30
Pool per task

vkhromov:
Allow specifying pool per task in addition or instead of specifying a pool for the executor.
If a task has a pool then the pool is used while matching an offer to the task, otherwise a pool from the executor is used while matching an offer to the task.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -215,15 +215,18 @@ class ExecutionFramework(Scheduler):
         self._reconcile_tasks_at += self._task_reconciliation_delay
 
     def offer_matches_pool(self, offer):
-        if self.pool is None:
-            # If pool is not specified, then we can accept offer from any agent
-            return True, None
+        offer_pool = None
 
         for attribute in offer.attributes:
             if attribute.name == "pool":
-                return attribute.text.value == self.pool, attribute.text.value
+                offer_pool = attribute.text.value
+                break
 
-        return False, None
+        if self.pool is None:
+            # If pool is not specified, then we can accept offer from any agent
+            return True, offer_pool
+
+        return offer_pool is not None and offer_pool == self.pool, offer_pool
 
     def kill_task(self, task_id):
         tmp_list = []
@@ -338,6 +341,9 @@ class ExecutionFramework(Scheduler):
             )
         )
 
+        offer_pool_match, offer_pool = self.offer_matches_pool(offer)
+        any_pool_match = False
+
         tasks_to_put_back_in_queue = []
 
         # Need to lock here even though we are working on the task_queue, since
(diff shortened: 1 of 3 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -215,15 +215,18 @@ def _reconcile_tasks(self, tasks_to_reconcile):
         self._reconcile_tasks_at += self._task_reconciliation_delay
 
     def offer_matches_pool(self, offer):
```
Rob-Johnson: I think we've got an opportunity to shuffle some bits round here.

Can you move this function to ``constraints.py`` and make a function ``pool_match`` just take a pool and offer (it's content can just be ``return equal_op(pool, offer_pool)``).

That way we don't have different logic for the executor pool and the offer pool (you use ``offer_matches_pool`` for the executor pool and an adhoc implementation on line 359.

Then we can have this function in the executor:
```
def offer_matches_pool(self, offer):
  if task.pool:
        return constraints.pool_match(attribute_pool, task.pool)
  elif self.pool:
         return constraints.pool_match(attribute_pool, self.pool)
  else:
        True
```

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -383,7 +396,16 @@ def get_tasks_to_launch(self, offer):
...
+        if not any_pool_match:
+            log.info(
+                "Declining offer {id}, its pool {op} matches "
+                "neither executor's pool {sp} nor any tasks' pool".format(
+                    id=offer.id.value,
+                    sp=self.pool,
+                    op=offer_pool
+                ))
+
+        return tasks_to_launch, any_pool_match
```
Rob-Johnson: I'd rather we just recalculated the pool match later on than return this tuple.

qui:
FYI - I don't mind having pool as a first-class config if there are other use cases for it, but as a part of the effort to separate Tron and PaaSTA, I'm planning on translating pool into generic constraints like ["pool", "LIKE", "default"] and removing the `pool` config from Tron. Looks like taskproc can already handle those types of constraints.

Here's a related PR: https://github.com/Yelp/paasta/pull/1812/

vkhromov:
So, by providing `pool=None` to `ExecutorFramework` ctr and `["pool", "==", "default"]` constraint to a task we could have the same result as after applying the diff from this pull.
Seems this pull should be abandoned.

Rob-Johnson:
@vkhromov you're right. in that case, maybe we should get rid of pool from the executor entirely, and just expect people to do the right thing with the task?

vkhromov:
@Rob-Johnson, yep, but first we need to switch existing users of `task_processing` to use pools in tasks, and only after that remove pool from the executor.

Rob-Johnson:
yep, ok.

PR #102, opened 2018-05-23
add global event invariant

keymone:
When event kind is 'task', we need to make a subset of attributes mandatory, which is not possible to do with per-field invariants.

### DIFF
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -16,7 +16,7 @@ EVENT_TASK_ATTRS = {'task_id', 'task_config'}
 class Event(PRecord):
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


task_processing/interfaces/event.py
```diff
@@ -9,10 +9,27 @@
...
 
-EVENT_KINDS = ['task', 'control']
+EVENT_KINDS = {'task', 'control'}
+EVENT_TASK_ATTRS = {'task_id', 'task_config'}
 
 
 class Event(PRecord):
+
+    @staticmethod
+    def validate_task(r):
```
Rob-Johnson: validate_task_event?

PR #104, opened 2018-06-06
Add option for logging executor to use python logging

qui:
I want to be able to direct messages for each task to a different file, and Python logging seems like the best way to do that. Here's an idea for how to add that functionality to the logging executor.

Very open to other ideas on how to achieve this. Maybe it would be better to create subclasses of LoggingExecutor that implement setup_loggers, etc?

### DIFF
--- a/task_processing/plugins/mesos/logging_executor.py
+++ b/task_processing/plugins/mesos/logging_executor.py
@@ -24,6 +24,7 @@ logging.getLogger("urllib3").setLevel(logging.WARNING)
 TASK_LOG_CHUNK_LEN = 4096
 DEFAULT_FORMAT = '{task_id}[{container_id}@{agent}]: {line}'
 
+
 class LogMetadata(PRecord):
     log_url = field(type=str, initial='')
     log_path = field(type=str, initial='')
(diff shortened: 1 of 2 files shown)


qui:
Thanks for the suggestion @keymone, that's much cleaner! Updated the code. Does that look like what you were imagining?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -34,9 +34,20 @@ class LogMetadata(PRecord):
...
-    def __init__(self, downstream_executor):
+    def __init__(
+        self,
+        downstream_executor,
+        handler=None,
+        format_string=None,
+    ):
         self.downstream_executor = downstream_executor
+        self.TASK_CONFIG_INTERFACE = downstream_executor.TASK_CONFIG_INTERFACE
+        self.handler = handler or standard_handler
```
keymone: this and format_string can just be defaults in the `def` right?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -57,6 +68,15 @@ def __init__(self, downstream_executor):
         self.logging_thread.daemon = True
         self.logging_thread.start()
 
+    def log_line(self, stream, line, task_id, container_id, agent):
+        formatted_line = self.format_string.format(
```
keymone: not a blocker, but maybe we don't want to pre-format the message? i'd leave that to the user.

qui: The user doesn't have the container_id or agent info, unless you want to pass those to the handler, too?

task_processing/plugins/mesos/logging_executor.py
```diff
@@ -34,9 +35,21 @@ class LogMetadata(PRecord):
...
 class MesosLoggingExecutor(TaskExecutor):
-    def __init__(self, downstream_executor):
+    def __init__(
+        self,
+        downstream_executor,
+        handler=standard_handler,
+        format_string=DEFAULT_FORMAT,
+    ):
         self.downstream_executor = downstream_executor
+        self.TASK_CONFIG_INTERFACE = downstream_executor.TASK_CONFIG_INTERFACE
```
Rob-Johnson: is this here so that the executor satisfies the interface?

qui: Yeah, I put this in because I'm getting the task interface from the runner: https://github.com/Yelp/Tron/blob/master/tron/mesos.py#L204

I could save the mesos task config interface instead of getting it from the runner. But this way, the stack's TASK_CONFIG_INTERFACE doesn't change when you add the logging executor, which I think makes more sense?

PR #105, opened 2018-06-07
clean up dev/tox/etc, strip out six (we only support 3.6 from now on)

drmorr0:
This commit strips out six from taskproc -- there's nothing that's using taskproc on Python2.7 and Python2.7 is basically EOLed anyways, so let's not bother.

I also clean up the development environment/tox/etc a little bit here.

### DIFF
--- a/examples/cluster/docker-compose.yaml
+++ b/examples/cluster/docker-compose.yaml
@@ -42,7 +42,6 @@ services:
       - mesosagent
       - dynamodb
     volumes:
-      - ../..:/src
       - /var/run/docker.sock:/var/run/docker.sock
   dynamodb:
     image: deangiberson/aws-dynamodb-local
--- a/examples/cluster/playground/Dockerfile
+++ b/examples/cluster/playground/Dockerfile
@@ -1,12 +1,17 @@
-FROM ubuntu:trusty
+FROM ubuntu:xenial
 
 RUN apt-get update -q && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        software-properties-common \
         debhelper dpkg-dev gcc gdebi-core git help2man libffi-dev \
         libssl-dev libsasl2-modules libyaml-dev pyflakes python3-dev python3-pip python3-pytest \
         python-tox python-yaml wget zip zsh \
         openssh-server docker.io curl vim jq libsvn-dev \
     && apt-get clean
+
+RUN add-apt-repository ppa:jonathonf/python-3.6 
+RUN apt-get update -q && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.6
 
 RUN cd /tmp && \
     wget http://mirrors.kernel.org/ubuntu/pool/universe/d/dh-virtualenv/dh-virtualenv_1.0-1_all.deb && \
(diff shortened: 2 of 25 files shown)


examples/cluster/docker-compose.yaml
```diff
@@ -43,7 +42,6 @@ services:
       - dynamodb
     volumes:
       - ../..:/src
-      - /var/run/docker.sock:/var/run/docker.sock
```
keymone: why is this removed? iirc it's needed for playground to be able to use host's docker.

examples/cluster/playground/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:trusty
+FROM ubuntu:bionic
```
keymone: this could cause to weird differences in behavior between example-cluster and what's running on stages/prod?

Rob-Johnson: yeah, can we just use xenial?

requirements.txt
```diff
@@ -1,5 +0,0 @@
-# Library so -e .
```
Rob-Johnson: why dont we need this anymore?

examples/cluster/docker-compose.yaml
drmorr0: Apparently this is no longer allowed?  I get this error when trying to start the playground:

```
tarting cluster_zookeeper_1 ... done
Starting cluster_dynamodb_1  ... done
Starting cluster_mesosmaster_1 ... done
Starting cluster_mesosagent_1  ... done
ERROR: Cannot create container for service playground: b"You're not allowed to mount /run/docker.sock in rw mode.\nYou are only allowed to bind mount things from these directories: [/nail/home/drmorr:rw /nail/bulkdata:ro /nail/etc:ro /nail/etc/srv-configs:ro /nail/sys/bin/indocker:ro /nail/live/private:ro /nail/srv:ro /nail/home/dockerbuild:ro /nail/scratch/drmorr:rw /scratch/drmorr:rw /etc/boto_cfg:ro /etc/boto.cfg:ro /etc/mrjob.conf:ro /etc/mrjob.cpu-intensive.conf:ro /etc/mrjob.memory-intensive.conf:ro /etc/mrjob.minimal.conf:ro /etc/ssh/ssh_known_hosts:ro /etc/yelp_clog.json:ro /var/run/synapse/services/:ro /var/run/synapse/sockets/:ro /run/synapse/services/:ro /run/synapse/sockets/:ro /etc/nodebot.conf:ro /etc/localtime:ro /usr/share/zoneinfo/:ro /etc/passwd:ro /etc/group:ro /nail/etc/zookeeper_discovery/:ro /nail/etc/kafka_discovery/:ro /nail/tmp/:rw /nail/opt/:ro /etc/send_nsca.cfg:ro /var/run/fusebox/:ro /run/fusebox/:ro]"
```

Rob-Johnson: thats a Yelp thing - you need to ``unset DOCKER_HOST``

requirements.txt
drmorr0: For libraries our recommended configuration is to put dependencies in setup.py and not have a requirements.txt:

> Libraries should not have deps = -rrequirements.txt in their tox files. Having a requirements.txt is not recommended in libraries.

https://yelpwiki.yelpcorp.com/display/WEBCORE/Applications+vs+Libraries

examples/cluster/docker-compose.yaml
drmorr0: aha thanks.

PR #106, opened 2018-06-13
pull out offers and resources into another file

drmorr0:
This change does the following:

* _actually_ use python3.6 in the examples Docker container; in my last request I just installed it, but all the scripts were still using python3.5
* bumps max line length to 100; there are so many long imports and things that we have to go through some pretty painful shenanigans to fit into 80 characters
* pulls out and DRYs up code around handling resource offers: this makes it more extensible and will get re-used when I create the pod executor
* changes the logic around how ports are handled; instead of passing around range() objects for all the ports that are available, we just pass a (begin, end) pair, which we update every time a port gets used
* some initial mypy support

### DIFF
--- a/task_processing/plugins/mesos/resource_helpers.py
+++ b/task_processing/plugins/mesos/resource_helpers.py
@@ -1,0 +1,91 @@
+from typing import Tuple
+
+from pyrsistent import field
+from pyrsistent import pmap
+from pyrsistent import PRecord
+from pyrsistent import PVector
+from pyrsistent import pvector
+from pyrsistent import v
+
+from task_processing.plugins.mesos.task_config import MesosTaskConfig
+
+NUMERIC_RESOURCE = field(
+    type=float,
+    initial=0.0,
+    factory=float,
+    invariant=lambda x: (x >= 0, 'resource < 0'),
+)
+_NUMERIC_RESOURCES = frozenset(['cpus', 'mem', 'disk', 'gpus'])
+
+
+class ResourceSet(PRecord):
+    cpus = NUMERIC_RESOURCE
+    mem = NUMERIC_RESOURCE
+    disk = NUMERIC_RESOURCE
+    gpus = NUMERIC_RESOURCE
+    ports = field(type=PVector, initial=v(), factory=pvector)
+
+
+def get_offer_resources(offer, role: str) -> ResourceSet:
+    """ Get the resources from a Mesos offer
+
+    :param offer: the payload from a Mesos resourceOffer call
+    :param role: the Mesos role we want to get resources for
+    :returns: a mapping from resource name -> available resources for the offer
+    """
+    res = ResourceSet()
+    for resource in offer.resources:
+        if resource.role != role:
+            continue
+
+        if resource.name in _NUMERIC_RESOURCES:
+            res = res.set(resource.name, resource.scalar.value)
+        elif resource.name == 'ports':
+            res = res.set('ports', [pmap(r) for r in resource.ranges.range])
+    return res
+
+
+def allocate_task_resources(
+    task: MesosTaskConfig,
+    offer_resources: ResourceSet,
+) -> Tuple[ResourceSet, ResourceSet]:
+    """ Allocate a task's resources to a Mesos offer
+
+    :param task: the specification for the task to allocate
+    :param offer_resources: a mapping of resource name -> available resources
+        (should come from :func:`get_offer_resources`)
+    :returns: a pair of (consumed_resources, remaining_resources)
+    """
+    consumed_resources = ResourceSet()
+    for res, val in offer_resources.items():
+        if res not in _NUMERIC_RESOURCES:
+            continue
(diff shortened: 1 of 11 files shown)


task_processing/plugins/mesos/resource_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    offer_resources = _make_resource_dict()
+    for resource in offer.resources:
+        if resource.role != role:
+            continue
+
+        if resource.name in NUMERIC_RESOURCES:
+            offer_resources[resource.name] += resource.scalar.value
+        elif resource.name == 'ports':
+            offer_resources['ports'] = [r for r in resource.ranges.range]
```
Rob-Johnson: why've you made the numeric_resources additive, wheras the ports are an assignment? I think, for a single offer, you'd expect exactly one resource for each of [``cpu``, ``mem``, ``disk``, ``gpu``] etc, so I don't think you need to do an additive thing? you can just set the value ``offer_resources[resource.name] += resource.scalar.value`` irrespective of whether it is a port or a numeric resource.

task_processing/plugins/mesos/resource_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+
+    return offer_resources
+
+
+def allocate_task_resources(task: MesosTaskConfig, offer_resources: Dict[str, Any]) -> int:
```
keymone: this function is a bit confusing: it mutates it's input and it returns a value that seems to be a single kind of "other" resource. i would prefer it to have a following form:

    consume_resources(
        requested_resources,
        offered_resources) -> (consumed_resources, remaining_resources)

where `consumed_resources` would mostly end up being a copy of `requested_resources` with special handling for `OTHER_RESOURCES` (like picking a port and whatnot) and `remaining_resources` would be used as an input in place of `offered_resources` for the next task. thoughts?

Rob-Johnson: I'd like to encourage you to not make this function rely on side effects, but just return a new dictionary that represents the offer_resource once the task config has been taken into account, and the caller can use that as they need to.

If we're going all in on the typing, can you also make the data structure that is returned by get_offer_resources something *real* (a named tuple, or a [PRecord](https://github.com/tobgu/pyrsistent#pr)) and update the type definition for the function. That way we can have more guarantees about the caller passing the right thing than those we get just asking them to read the docstring.

Finally it feels awkward that this returns a port (seemingly completely unrelated to whats hinted at by the function name). Can we separate the function that provides a new offer_resources with the numeric resources subtracted from that which knows how to deal with ports?

Rob-Johnson: btw, I kind of hinted at it, but this is exactly the kind of pattern that pyrsistent forces out of your code for you.You couldn't write this function using a immutable data structure for offer_resources.

drmorr0: sgtm!

task_processing/plugins/mesos/resource_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+
+from task_processing.plugins.mesos.task_config import MesosTaskConfig
+
+NUMERIC_RESOURCES = frozenset(['cpus', 'mem', 'disk', 'gpus'])
+OTHER_RESOURCES: dict = {'ports': []}
+
+
+def _make_resource_dict():
+    d = dict([(r, 0) for r in NUMERIC_RESOURCES])
+    d.update(OTHER_RESOURCES)
```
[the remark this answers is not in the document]
drmorr0: huh good catch.

task_processing/plugins/mesos/resource_helpers.py
```diff
@@ -0,0 +1,69 @@
+from typing import Any
+from typing import Dict
+
+from task_processing.plugins.mesos.task_config import MesosTaskConfig
+
+NUMERIC_RESOURCES = frozenset(['cpus', 'mem', 'disk', 'gpus'])
```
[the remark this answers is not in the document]
drmorr0: mypy didn't complain, but I can

task_processing/plugins/mesos/resource_helpers.py
drmorr0: Fixed, see maksym's comment above.

task_processing/plugins/mesos/resource_helpers.py
drmorr0: discussed on slack, fixed

task_processing/plugins/mesos/resource_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+OTHER_RESOURCES: dict = {'ports': []}
+
+
+def _make_resource_dict():
```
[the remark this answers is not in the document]
drmorr0: fixed

drmorr0:
From robj: 

> what do you think about making the Dict[str, Any] that you return from get_offer_resources a real type?
> fix and ship

PR #110, opened 2018-06-25
minor bug fixes

drmorr0:


examples/async.py
```diff
@@ -50,7 +50,7 @@ def main():
         task_config = TaskConfig(image='busybox', cmd='/bin/true')
         runner.run(task_config)
 
-    for _ in range(5):
```
sagar8192: I think is intentionally set to 5 because we want to wait for 10 secs.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -530,10 +530,7 @@ def statusUpdate(self, driver, update) -> None:
         task_id = update.task_id.value
         task_state = str(update.state)
 
-        log.info("Task update {update} received for task {task}".format(
-            update=task_state,
-            task=task_id
-        ))
+        log.info(f'Task update {task_state} received for task {task_id} at time {update.timestamp}')
```
sagar8192: hmm... update.timestamp is when the update was generated. Maybe reword this or change it to time.time

PR #112, opened 2018-07-05
fix events without raw field

keymone:
- make .raw attribute mandatory
- include .raw in offer timeout terminal event
- also reformat some code for readability

### DIFF
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -14,21 +14,15 @@ EVENT_TASK_ATTRS = {'task_id', 'task_config'}
 
 
 class Event(PRecord):
-
-    @staticmethod
-    def validate_task_event(r):
-        if r.kind != 'task':
+    def __invariant__(self):
+        if self.kind != 'task':
             return True, 'Not a task event'
 
-        missing_attrs = EVENT_TASK_ATTRS - set(r.keys())
+        missing_attrs = EVENT_TASK_ATTRS - set(self.keys())
         if missing_attrs:
             return False, 'Missing task attributes: {}'.format(missing_attrs)
 
         return True, 'Task is valid'
-
-    def __invariant__(r): return (
-        Event.validate_task_event(r),
-    )
 
     kind = field(type=str,
                  mandatory=True,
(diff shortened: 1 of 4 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -140,11 +135,13 @@ def _background_check(self):
                                 task_event(
                                     task_id=task_id,
                                     terminal=True,
-                                    timestamp=time.time(),
+                                    timestamp=time_now,
```
chlgit: It might not make any difference because the granularity is second. But I think time.time() is more accurate than time_now?

task_processing/interfaces/event.py
```diff
@@ -38,7 +31,7 @@ def __invariant__(r): return (
     # use time.time() to generate
     timestamp = field(type=float)
     # reference to platform-specific event object
-    raw = field(initial=addict.Dict())
+    raw = field(mandatory=True)
```
sagar8192: I think the raw field is supposed to be of type addict.Dict. Can you add the type here?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -140,11 +135,13 @@ def _background_check(self):
...
                                     success=False,
                                     message='stop',
                                     task_config=md.task_config,
-                                ))
+                                    raw='Failed due to offer timeout',
```
sagar8192: Can you change this to addict.Dict to make it consistent with other places?

keymone: as this field is dedicated for platform-specific event data it can end up being any datatype, depending on the backend or even mesos library version. we shouldn't give anyone ideas that they can rely on form of this value.

task_processing/plugins/mesos/execution_framework.py
keymone: the whole loop should be done in milliseconds while involved timeout are expressed in hundreds/thousands of seconds, cached `time.time()` call is fine here. besides, `time_now` is what the logic of conditionals is based on - it makes sense to report it with any resulting events.

task_processing/interfaces/event.py
keymone: sorry, didn't notice this comment before replying to the other one: https://github.com/Yelp/task_processing/pull/112/files/e83d65482c875803a43536525d4b1d393ad0e028#r200470432

PR #113, opened 2018-07-23
remove logging from task_processor.py

drmorr0:


task_processing/task_processor.py
```diff
@@ -121,8 +117,11 @@ def executor_cls(self, provider):
         if provider in self.registry.task_executors:
             return self.registry.task_executors[provider]
         elif provider in self.registry.deprecated_task_executors:
-            log.warning(
-                f'{provider} is a deprecated executor and will be removed in the future')
+            FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(funcName)s - %(message)s'
+            logging.basicConfig(format=FORMAT)
+            log = logging.getLogger(__name__)
```
qui: What if we kept this on the module level and took out the basicConfig entirely? Since basicConfig can only be called once, I think we should leave it up to the application in general.

(I would be in favor of removing the basicConfig calls from plugins/mesos/, even if they don't necessarily break things on import, but that could affect current users if they depend on the format being set this way.)

drmorr0: Yea, I was thinking about this too; I'll make that change.

PR #116, opened 2018-08-09
Always suppress incoming offers when task queue is empty

huadongliu:
This is to deal with leader re-election. It does not hurt on normal conditions.

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -423,16 +423,17 @@ class ExecutionFramework(Scheduler):
 
         with self._lock:
             if self.task_queue.empty():
-                if not self.are_offers_suppressed:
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -423,16 +423,17 @@ def resourceOffers(self, driver, offers) -> None:
 
         with self._lock:
             if self.task_queue.empty():
-                if not self.are_offers_suppressed:
```
sagar8192: You can get rid of this variable now, right?

huadongliu: are_offers_suppressed is still used in enqueue_task.

sagar8192: Ah! ship-it

keymone:
do we not get any notification from mesos that leader election happened?

PR #117, opened 2018-08-10
Enable re-registering with an existing framework ID

qui:


chlgit:
I've updated this at #119 and merged into master. Will close this PR.

PR #120, opened 2018-08-23
[WIP] Add check method to executor that does Mesos reconciliation

qui:
Calling it "check" instead of reconcile or something because I'm trying to make it plugin-agnostic...

I'm conflicted about taking a task_config as an argument to check, instead of a task_id. On one hand, seems like task_id should be sufficient to check the status of a task. But then we'd need to do `task_config=m()` when we create the TaskMetadata in recover_task, and other places might assume that task_config isn't empty. Maybe one solution is to create a different data structure besides `task_metadata` for storing these recovery tasks? Let me know what you think.


task_processing/plugins/mesos/task_config.py
```diff
@@ -118,7 +118,10 @@ def __invariant__(conf):
         invariant=_valid_constraints,
     )
     use_cached_image = field(type=bool, initial=True, factory=bool)
+    task_id_override = field(type=(str, type(None)), initial=None)
```
keymone: where do we use this?

qui:
Actually, changed my mind about returning something, I'll call it reconcile instead.

qui:
Sure, we don't really need task_id_override, we can construct it using the uuid instead. Took it out.

keymone:
i think it would make sense to provide task_id setter for config classes that have special rule on how they construct it. but it's not a blocker.

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -199,6 +199,33 @@ def _background_check(self):
             )
             time.sleep(10)
 
+    def reconcile_task(self, task_config):
```
Rob-Johnson: mesos supports 'implicit' reconciliation, when you can call `reconcile` with an empty list of task ids and it returns everything. do we want to support that?

qui: Sure, we could let `task_config` be optional in all the `reconcile` methods and do implicit reconciliation in the framework if the config is None. 

That kind of makes the `reconcile` method more specific to Mesos, but if we think it would be useful to have implicit reconciliation, I don't think that matters too much.

qui: Just remembered the framework currently discards status updates for tasks that don't exist in task_metadata, so we'd have to decide how to deal with that, too. Since I wasn't planning on using implicit reconciliation in Tron for now, how about leaving this until we decide we need it?

PR #124, opened 2018-09-07
Don't raise exception if running task is missing

qui:
The task_id may not be in `self.staging_tasks` if we have restarted and are receiving reconciliation events. We want the thread to continue processing events. Also added some tests for the logging executor.

I wanted to fetch logs again, but given the TASK_RUNNING update only has the agent_id, it doesn't seem possible to get the agent_url back. (We don't have the master url anywhere, to get the agent url from the agent id.) This is an example of the update:
```
{
'agent_id': {'value': '2bd4b611-9206-47ca-a64d-7b7a2973b9e0-S1'}, 
'container_status': {
  'container_id': {'value': '...'}, 
  'network_infos': [{
      'ip_addresses': [{'ip_address': '169.254.0.1', 'protocol': 'IPv4'}]
    }]
},
'data': '...', 
'executor_id': {'value': 'MASTER.test.1.first.930ec85b-10a5-43e9-8970-05fee5a1d247'}, 
'labels': {'labels': [{'key': 'Docker.NetworkSettings.IPAddress', 'value': '169.254.0.1'}]}, 
'source': 'SOURCE_EXECUTOR',
'state': 'TASK_RUNNING',
'task_id': {'value': 'MASTER.test.1.first.930ec85b-10a5-43e9-8970-05fee5a1d247'},
'timestamp': 1536190686.70458,
'uuid': '/IHnK8J5QuGNggG0uolaDA=='
}
```

### DIFF
--- a/task_processing/plugins/mesos/logging_executor.py
+++ b/task_processing/plugins/mesos/logging_executor.py
@@ -171,6 +171,10 @@ class MesosLoggingExecutor(TaskExecutor):
                     self.staging_tasks = self.staging_tasks.set(e.task_id, url)
 
                 if e.kind == 'task' and e.platform_type == 'running':
+                    if e.task_id not in self.staging_tasks:
+                        log.info(f"Task {e.task_id} already running, not fetching logs")
+                        continue
+
                     url = self.staging_tasks[e.task_id]
                     self.staging_tasks = self.staging_tasks.discard(e.task_id)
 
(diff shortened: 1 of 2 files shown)


task_processing/plugins/mesos/logging_executor.py
```diff
@@ -171,6 +171,10 @@ def event_loop(self):
                     self.staging_tasks = self.staging_tasks.set(e.task_id, url)
 
                 if e.kind == 'task' and e.platform_type == 'running':
+                    if e.task_id not in self.staging_tasks:
+                        log.warning(f"Not tracking logs for task {e.task_id}")
```
keymone: i would make this a debug statement "Task already running" or something, recovery and reconciliation should be "normal" events, not something that triggers warnings

PR #125, opened 2018-09-17
Modify timeout and retrying executors to allow missing tasks to reregister during reconciliation

kawaiwanyelp:
### Description
- Update timeout and retrying executors to allow missing tasks to reregister during reconciliation
- For the timeout executor, if it now encounters a non-terminal task update that the executor doesn't know about, it will add the task to its list of running_tasks using the update's timestamp as a baseline from which to establish a timeout deadline.
- For the retrying executor, if it now encounters a task update that the executor doesn't know about, it will similarly add the task to its list of task_retries and assumes the update's attempt number is the number of attempts so far.
- The above 2 points are a consequence of needing to pull information that is lost unless we can somehow recover task histories (like when a task started).

### Testing done
- make test

### DIFF
--- a/task_processing/plugins/mesos/timeout_executor.py
+++ b/task_processing/plugins/mesos/timeout_executor.py
@@ -37,16 +37,29 @@ class TimeoutExecutor(TaskExecutor):
                 e = self.src_queue.get()
                 self.dest_queue.put(e)
 
-                if not (e.kind == 'task' and e.terminal):
+                if not e.kind == 'task':
                     continue
-                # Update running and killed tasks
-                with self.tasks_lock:
-                    for idx, entry in enumerate(self.running_tasks):
-                        if e.task_id == entry.task_id:
-                            self.running_tasks.pop(idx)
-                            break
-                    if e.task_id in self.killed_tasks:
-                        self.killed_tasks.remove(e.task_id)
+                elif not e.terminal:
+                    with self.tasks_lock:
+                        if e.task_id not in [entry.task_id for entry in self.running_tasks]:
+                            # No record of e's task_id in self.running_tasks,
+                            # so we need to add it back in. We lack access to
+                            # the original time the task was started, so to set
+                            # a deadline, we use e's timestamp as a baseline.
+                            new_entry = TaskEntry(
+                                task_id=e.task_id,
+                                deadline=e.task_config.timeout + e.timestamp,
+                            )
+                            self._insert_new_running_task_entry(new_entry)
+                else:
+                    # Update running and killed tasks
+                    with self.tasks_lock:
+                        for idx, entry in enumerate(self.running_tasks):
+                            if e.task_id == entry.task_id:
+                                self.running_tasks.pop(idx)
+                                break
+                        if e.task_id in self.killed_tasks:
+                            self.killed_tasks.remove(e.task_id)
 
             # Check timeouts
             current_time = time.time()
@@ -83,11 +96,7 @@ class TimeoutExecutor(TaskExecutor):
             deadline=task_config.timeout + time.time()
         )
         with self.tasks_lock:
-            for idx, entry in enumerate(self.running_tasks):
-                if new_entry.deadline <= entry.deadline:
-                    self.running_tasks.insert(idx, new_entry)
-                    return
-            self.running_tasks.append(new_entry)
+            self._insert_new_running_task_entry(new_entry)
 
         self.downstream_executor.run(task_config)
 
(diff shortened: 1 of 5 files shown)


tests/unit/plugins/mesos/retrying_executor_test.py
```diff
@@ -59,37 +73,169 @@ def test_task_retry(mock_retrying_executor):
...
-    mock_task_config = _get_mock_task_config()
+def test_retry_loop_retries_task(mock_retrying_executor, mock_event):
+    mock_event = mock_event.set('terminal', True)
+    mock_retrying_executor.stopping = True
+    mock_retrying_executor._is_current_attempt = mock.Mock(return_value=True)
+    mock_retrying_executor._restore_task_id = mock.Mock(
+        return_value=mock_event)
+    mock_retrying_executor.retry = mock.Mock(return_value=True)
+    mock_retrying_executor.retry_pred = mock.Mock(return_value=True)
+    mock_retrying_executor.src_queue = Queue()
```
qui: You can use the fixture instead.

tests/unit/plugins/mesos/retrying_executor_test.py
```diff
@@ -59,37 +73,169 @@ def test_task_retry(mock_retrying_executor):
...
+    mock_retrying_executor._is_current_attempt = mock.Mock(return_value=True)
+    mock_retrying_executor.event_with_retries = mock.Mock()
+    mock_retrying_executor.retry = mock.Mock()
+    mock_retrying_executor.src_queue = Queue()
+    mock_retrying_executor.src_queue.put(mock_event)
+
+    mock_retrying_executor.retry_loop()
+
+    assert mock_retrying_executor.dest_queue.qsize() == 1
+    assert mock_retrying_executor.retry.call_count == 0
```
qui: I think it's a little overkill to assert this, the code would definitely break before then because control events don't have task ids, etc. It's enough to know that the task got passed on to dest_queue and the loop didn't raise an exception. That way you don't need as much setup code, which would need updating in the future.

tests/unit/plugins/mesos/timeout_executor_test.py
```diff
@@ -13,15 +17,136 @@ def mock_Thread():
...
+    mock_timeout_executor.stopping = True
+    mock_timeout_executor.src_queue = Queue()
+    mock_timeout_executor.src_queue.put(mock_event)
+    mock_timeout_executor.running_tasks.append(mock_entry)
+    mock_timeout_executor.downstream_executor.kill = mock.Mock()
+    time.time = mock.Mock(return_value=0)
+
+    mock_timeout_executor.timeout_loop()
+
+    assert mock_timeout_executor.downstream_executor.kill.call_count == 0
```
qui: Again, we should make this simpler and just assert the control event is on the queue.

tests/unit/plugins/mesos/timeout_executor_test.py
```diff
@@ -13,15 +17,136 @@ def mock_Thread():
...
+
+# timeout_loop #################################################################
+def test_timeout_loop_nontask(
+    mock_timeout_executor,
+    mock_event,
+):
+    mock_event = mock_event.set('kind', 'control')
+    mock_entry = TaskEntry('different_id', deadline=1234)
+    mock_timeout_executor.stopping = True
+    mock_timeout_executor.src_queue = Queue()
```
qui: Let's use the fixture everywhere.

task_processing/plugins/mesos/timeout_executor.py
```diff
@@ -83,11 +96,7 @@ def run(self, task_config):
             deadline=task_config.timeout + time.time()
         )
         with self.tasks_lock:
-            for idx, entry in enumerate(self.running_tasks):
-                if new_entry.deadline <= entry.deadline:
-                    self.running_tasks.insert(idx, new_entry)
-                    return
```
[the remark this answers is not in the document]
kawaiwanyelp: My initial thought was that we wanted to run no matter what, but looking over it again I see that it's to prevent re-running duplicated tasks. I'll fix it to return and prevent run() from executing, like before.

tests/unit/plugins/mesos/retrying_executor_test.py
kawaiwanyelp: Fair.

tests/unit/plugins/mesos/retrying_executor_test.py
kawaiwanyelp: Sure.

tests/unit/plugins/mesos/timeout_executor_test.py
kawaiwanyelp: Sure.

tests/unit/plugins/mesos/timeout_executor_test.py
kawaiwanyelp: Sure.,

PR #128, opened 2018-10-01
Tasks accounting fixes

keymone:
- set task state to TASK_STUCK to track that state separately
- log stuck tasks that didn't receive terminal update explicitly
- after every hour spent in  TASK_STUCK, send another kill request
- log and ignore task configs we're about to launch that are missing from
metadata
- decline offer when launching tasks fails
- correctly decline offer when no tasks were launched


keymone:
updated PR to retry killing stuck tasks once an hour

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -170,29 +173,47 @@ def _background_check(self):
...
+                                    'TASK_STUCK', time.time()),
+                            )
+                        )
                         self.blacklist_slave(
                             agent_id=self.task_metadata[task_id].agent_id,
                             timeout=self.slave_blacklist_timeout_s,
                         )
                         get_metric(metrics.TASK_STUCK_COUNT).count(1)
 
+                    if md.task_state == 'TASK_STUCK':
```
Rob-Johnson: can you just add a comment to give us a high level overview of what this does please?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -170,29 +173,47 @@ def _background_check(self):
...
+                        t = time.time()
+                        # account for time we spent in this loop
+                        time_delta = t - time_now + 10
```
keymone: to clarify this logic:

`time_delta` - 10 seconds we slept since the last iteration of background check started + seconds we spent in current iteration
`time_stuck` - seconds since task was put in TASK_STUCK state
`hour_rolled` - seconds since `time_stuck` crossed another hour boundary

basically we want to send kill request in the first 0..10s interval of each hour. there are still some unlikely timings when we will miss retrying a kill, but i really don't want to maintain any more state in this framework.

task_processing/plugins/mesos/execution_framework.py
keymone: done

keymone:
@sagar8192 we only re-enqueue tasks from UNKNOWN state and when we do that, state is set to TASK_INITED. Also when terminal state is received - task metadata is discarded no matter which state it is in.

What do you think should happen when  we get terminal update for stuck task?

PR #129, opened 2018-10-01
Split execution framework background check

keymone:
- move per-task logic out of `_background_check` into `_background_check_task`
- log timing information for each bg check run

### DIFF
--- a/task_processing/plugins/mesos/execution_framework.py
+++ b/task_processing/plugins/mesos/execution_framework.py
@@ -122,6 +122,95 @@ class ExecutionFramework(Scheduler):
             log.warning('{} failed: {}'.format(method, str(e)))
             return self.driver_error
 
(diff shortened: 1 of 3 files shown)


task_processing/plugins/mesos/execution_framework.py
```diff
@@ -131,101 +220,18 @@ def _background_check(self):
...
+                        tasks_to_reconcile,
+                        task_id,
+                        md,
+                    )
 
             self._reconcile_tasks(
                 [Dict({'task_id': Dict({'value': task_id})}) for
                     task_id in tasks_to_reconcile]
             )
+            log.info(f'background check done in {time.time() - time_now}s')
```
sagar8192: +1 This is nice. Can you also add a metric for this? We can create a dashboard to track this.

keymone: good point! done

PR #130, opened 2018-10-09
Fix incorrect log message when offer is declined and missing metric in execution framework

keymone:
- no `continue` near :616 would result in both declined and accepted log messages
- create missing timer metric (line 434)
- change formatting to f-strings (the rest)


qui:
Test apparently needs to be updated, though.

PR #132, opened 2018-10-31
Only get_metric for insufficient offer once

qui:
tzhu made this suggestion, apparently it improves jolt cpu usage by 10%:
```
status quo: 32:32min CPU in 48min (0.677 cpu)
optimized: 8:00~8:15min CPU in 15min (0.541 cpu)
```
I checked the other places we call get_metric, nothing else obvious to optimize.


qui:
I think it's probably particularly expensive for jolt because they start with a large number of pending tasks.

PR #134, opened 2018-12-18
Stop dynamodb persister from persisting control events

kawaiwanyelp:
### Description
- Currently, the DynamoDB persister persists both task and control events. The problem is that control events do not `task_id`s, but our DynamoDB tables expect them.
- In this change, I made it so that we do not persist control events, only task events.

### Testing
- manual testing on a dev paasta master
- make test

### Notes to reviewers
- Error when persisting control events: https://gist.github.com/kawaiwanyelp/886819f047a1785e3e823a8771100ff6
- This change also includes a bunch of changes to silence or fix mypy errors. 
    - pyrsistent seems to be bugged in that `PMap`s and `PVector`s do not seem to play well with mypy. Hence, I've instructed `mypy` to ignore them using `# type: ignore`

### DIFF
--- a/task_processing/interfaces/event.py
+++ b/task_processing/interfaces/event.py
@@ -1,10 +1,11 @@
 import json
 import uuid
+from typing import Any
 
 from pyrsistent import field
 from pyrsistent import freeze
 from pyrsistent import m
-from pyrsistent import PMap
+from pyrsistent import PMap  # type: ignore
 from pyrsistent import pmap
 from pyrsistent import PRecord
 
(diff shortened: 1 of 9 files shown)


Rob-Johnson:
in theory the type annotations exist in v0.14.7 of pyrsistent - are we using an updated version?

kawaiwanyelp:
@Rob-Johnson We are using the latest version.

I've checked the type annotations, and it seems like they're set incorrectly, or at least not in a way that plays well with mypy. For example, mypy complains that there is no attribute `PMap` in the pyrsistent module, even though there is. If I use `pyrsistent._pmap.PMap` (that's what it really is), mypy complains about something else. I could import `pyrsistent.typing.PMap`, but that isn't the same as the real PMap, and that brings up another typing issue.

I don't think its worth putting a ton of effort into this so I think ignoring these specific cases is fine.

PR #137, opened 2019-03-22
Restart driver if it exits for any reason

qui:
The driver exits on things like HTTP error codes or parsing errors. I think it makes sense to reconnect in these cases, instead of letting the execution framework continue when the driver thread has stopped.

example of the driver exiting: https://github.com/douban/pymesos/blob/master/pymesos/process.py#L162

### DIFF
--- a/task_processing/plugins/mesos/mesos_executor.py
+++ b/task_processing/plugins/mesos/mesos_executor.py
@@ -79,10 +79,16 @@ class MesosExecutor(TaskExecutor):
         )
 
         # start driver thread immediately
+        self.stopping = False
         self.driver_thread = threading.Thread(
-            target=self.driver.run, args=())
+            target=self._run_driver, args=())
         self.driver_thread.daemon = True
         self.driver_thread.start()
+
+    def _run_driver(self):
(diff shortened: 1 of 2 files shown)


qui:
Tested this by adding an exception to the driver process, here's the log output:
```
2019-03-21 15:59:30,176 pymesos.process ERROR Thread abort:
Traceback (most recent call last):
  File "/nail/home/qui/work/Tron/.tox/py36/lib/python3.6/site-packages/pymesos/process.py", line 374, in _run
    if not conn.read():
  File "/nail/home/qui/work/Tron/.tox/py36/lib/python3.6/site-packages/pymesos/process.py", line 105, in read
    raise Exception('random error')
Exception: random error
2019-03-21 15:59:30,176 task_processing.plugins.mesos.mesos_executor WARNING Driver stopped, starting again
2019-03-21 15:59:30,178 task_processing.plugins.mesos.execution_framework WARNING Re-registered to {'hostname': '10-40-11-43-uswest1adevc.dev.yelpcorp.com', 'port': 5050, 'version': '1.7.2'} with role *
```

PR #138, opened 2019-07-31
Make the logging_executor tolerant of weird offers

solarkennedy:


task_processing/plugins/mesos/logging_executor.py
```diff
@@ -165,9 +165,16 @@ def event_loop(self):
...
+                    try:
+                        url = e.raw.offer.url.scheme + '://' + \
+                            e.raw.offer.url.address.ip + ':' + \
+                            str(e.raw.offer.url.address.port)
+                    except Exception as exc:
+                        log.error(
+                            f"Error decoding the url for this offer: {e.raw.offer.url}. "
+                            f"Setting to None. Exception: {exc}"
+                        )
+                        url = None
```
qui: Hm, in the traceback I found, the error occurred on the reconciliation event from the background check - so I think we already had a url at that point. Can you check if the task id is already in `self.staging_tasks` and skip the update if so?

solarkennedy: Hmm. The stack trace I found was *definitely* here. 
Maybe we have more than one bug!? :)

Can you paste exactly what you saw?
That may be a different bug, but I will also do what you asked where we skip the update if the staging tasks already has this taskid

qui: I'm talking about the same stack trace: https://fluffy.yelpcorp.com/i/fV4wBLpP6JGGgCdSN62JCmKV545FsHGl.html

I didn't mean it was in a different place, I meant the event that caused it was a reconciliation event (which you can see from the log lines above).

PR #140, opened 2019-08-01
Try to extract the log_url from offers in a more robust way

solarkennedy:


task_processing/plugins/mesos/logging_executor.py
```diff
@@ -246,3 +247,17 @@ def reconcile(self, task_config):
...
+
+
+def extract_url_from_offer(e):
+    try:
+        url = e.raw.offer.url.scheme + '://' + \
+            e.raw.offer.url.address.ip + ':' + \
+            str(e.raw.offer.url.address.port)
+    except Exception as exc:
+        log.error(
+            f"Error decoding the url for this offer: {e.raw.offer.url}. "
```
qui: Are you sure that RUNNING task updates have an offer? I didn't think they did... Either way, we should just log `e.raw.get('offer')` instead of assuming too much in this exception handler.

solarkennedy: Hmm. No you are probably right, was assuming. Hmm.
I'm just worried that if we get "None" early on, that we won't be able to get logs from a task.
I guess we are just crashing now, just getting no logs is better than crashing.

solarkennedy:
@qui can you look at this again to see if I've properly addressed your feedback?

PR #143, opened 2020-05-05
Make the LIKE constraint a "full match"

drmorr0:
(COMPINFRA-333) We have a "stable" pool and a "stable_batch" pool, and
jobs that are supposed to run on the "stable" pool are getting scheduled
on the "stable_batch" pool because the constraint is too open. This
change should make it so that stuff only gets scheduled on the pool they
specify.

(deleted coverage because it's broken in Python 3.6.0)


qui:
Can you add a test for this?

drmorr0:
> Can you add a test for this?

Sure, added a test and confirmed that the test fails on master.

PR #144, opened 2020-08-04
Add OWNERS file

mattmb:
So other teams at Yelp know who to hold responsible

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,3 @@
+---
+teams:
+- Compute Infra <compute-infra@yelp.com>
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -0,0 +1,3 @@
+---
+teams:
+- Compute Infra <compute-infra@yelp.com>
```
nemacysts, line 3: we can probably merge this after a small change:
```suggestion
- Compute Infrastructure Platform Experience
```

(we no longer require emails in OWNERS files internally so we can omit it and potentially save ourselves some spam should we ever allow these address to receive external mail - not that these being well-formed matters since we're not indexing our OSS yet, tho :p)

keymone (approved):

PR #147, opened 2021-05-27
TASKPROC-229 - Create Kubernetes Version of DefaultTaskConfigInterface

lanceleih:
### Description
Creating a base on what the Kubernetes plugin will require to start a Pod.


keymone (approved):

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+                False,
+                'Invalid volume format, must only contain following keys: '
+                '{}, was: {}'.format(VOLUME_KEYS, vol.keys())
+            )
+    return (True, None)
+
+class KubernetesTaskConfig(DefaultTaskConfigInterface):
+    uuid = field(type=(str, uuid.UUID), initial=uuid.uuid4)
+    node_name = field(type=str)
+    node_selector = field(type=(str,str))
```
nemacysts, line 30: I think we can probably want to store this in a closer form to how k8s does, so this should be a [PMap](https://pypi.org/project/pyrsistent/#pmap) (unless you see an opportunity to do some validation here, in which case we'd want a PRecord)

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+    containers = field(type=PVector)
+    restart_policy = field(type=str)
```
nemacysts, line 32: hmm, can we do an enum here? there's a finite set of restart policies, so it'd be nice to have a more ergonomic way of setting a restart policy than just passing in arbitrary strings (since they're easy to typo)

I think we can also set a default of `Never` since I think we can assume that task_proc users (at least internally) want to signal that their tasks are idempotent by explicitly setting a restart policy themselves - and both Tron (and Jolt, iirc) handle retries themselves anyway so it'd be nicer for us too :p

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+    # ToDO IAM stuff
+    service_account_name = field(type=str)
```
nemacysts, line 34: we might wanna circle back to this (or see how PaaSTA sets this) - do we get this for free if we use a kubeconfig file for setting ourselves up or does paasta also put something in here?

task_processing/plugins/kubernetes/task_config.py
nemacysts: or actually, maybe just hardcode to Never for now? I think it's not worth implementing any retry/restart stuff in task_proc just yet since our initial user (Tron) won't use it - I think Jolt does, but we can cross that bridge when we get there since it's not yet clear what Jolt on k8s will look like

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+        if set(vol.keys()) != VOLUME_KEYS:
+            return (
+                False,
+                'Invalid volume format, must only contain following keys: '
+                '{}, was: {}'.format(VOLUME_KEYS, vol.keys())
+            )
+    return (True, None)
+
+class KubernetesTaskConfig(DefaultTaskConfigInterface):
+    uuid = field(type=(str, uuid.UUID), initial=uuid.uuid4)
```
nemacysts, line 28: we can probably omit this and just use the inherited uuid from the superclass if we ever need a uuid, right?

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+    @property
+    def task_id(self):
+        return "{}.{}".format(self.name, self.uuid)
+
+    def set_task_id(self, task_id):
+        try:
+            name, uuid = task_id.rsplit('.', maxsplit=1)
+        except ValueError:
+            raise ValueError(f'Invalid format for task_id {task_id}')
+        return self.set(name=name, uuid=uuid)
```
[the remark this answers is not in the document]
nemacysts: (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names are the restrictions for a pod name if we want to add any validation)

task_processing/plugins/kubernetes/task_config.py
nemacysts: oh, looks like in paasta we only use this for operators? we can probably remove this then :)

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+
+def valid_volumes(volumes):
+    for vol in volumes:
+        if set(vol.keys()) != VOLUME_KEYS:
+            return (
+                False,
+                'Invalid volume format, must only contain following keys: '
+                '{}, was: {}'.format(VOLUME_KEYS, vol.keys())
+            )
+    return (True, None)
```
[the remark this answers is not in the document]
lanceleih: Removed.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+
+class KubernetesTaskConfig(DefaultTaskConfigInterface):
+    uuid = field(type=(str, uuid.UUID), initial=uuid.uuid4)
+    node_name = field(type=str)
+    node_selector = field(type=(str,str))
+    containers = field(type=PVector)
+    restart_policy = field(type=str)
+    # ToDO IAM stuff
```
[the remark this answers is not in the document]
lanceleih: We don't have a ticket for it yet. I think it makes sense for us to add IAM related work inside the TaskConfig, but I will hold onto that until we discuss further.

For now, I will remove the comment.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Removed.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+    # overrides the executor setting.
+    retries = field(type=int,
+                    factory=int,
+                    mandatory=False,
+                    invariant=lambda r: (r >= 0, 'retries >= 0'))
+    volumes = field(type=PVector,)
+
+    @property
+    def task_id(self):
+        return "{}.{}".format(self.name, self.uuid)
```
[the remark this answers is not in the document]
lanceleih: I think `name` is good to keep for now because it's one of the properties inside DefaultTaskConfigInterface. Changing that `pod_name` doesn't really make it too general for other plugins to be created from this interface.

task_processing/plugins/kubernetes/task_config.py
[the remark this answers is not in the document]
lanceleih: Added validation and type annotation.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,52 @@
...
+            return (
+                False,
+                'Invalid volume format, must only contain following keys: '
+                '{}, was: {}'.format(VOLUME_KEYS, vol.keys())
+            )
+    return (True, None)
+
+class KubernetesTaskConfig(DefaultTaskConfigInterface):
+    uuid = field(type=(str, uuid.UUID), initial=uuid.uuid4)
+    node_name = field(type=str)
```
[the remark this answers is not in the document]
lanceleih: Removed.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Omitted.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Changed to PMap.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Done.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @property
+    def pod_name(self) -> str:
+        return self.name
+
+    def set_pod_name(self, pod_name: str):
+        try:
+            name, uuid = pod_name.rsplit('.', maxsplit=1)
+        except ValueError:
+            raise ValueError(f'Invalid format for pod_name {pod_name}')
```
nemacysts, line 31: we'll need to figure out what to do with the uuid (or just ignore it) at some point, but we can probably leave this for now until we're actually working on handling tron restarts (since right now it's not particularly clear that you need to have saved the previous uuid if you want to reconstruct a task config and make it match exactly to the previous one)

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,43 @@
...
+
+        if len(pod_name) > 253:
```
nemacysts, line 33: can we make a named constant for this?
```suggestion
        MAX_POD_NAME_LENGTH = 253
        ...
        if len(pod_name) > MAX_POD_NAME_LENGTH:
```

nemacysts (commented):
looking good - just some minor nits in the `set_pod_name` function! I do wonder if it's worth trying to do the length/regex check as a pyrsistent `invariant` (e.g., like how retries are validated) but I'll leave that up to you :)

task_processing/plugins/kubernetes/task_config.py
keymone: the point of uuid is to have unique task name, in mesos plugin it's generated automatically upon creating task object: https://github.com/Yelp/task_processing/blob/master/task_processing/plugins/mesos/task_config.py#L60, in kubernetes we probably want to do the same as pod names should also be unique.

nemacysts: yea - I was thinking that we could do whatever kubernetes does internally to prevent duplicates (since I remember reading something about how k8s handles conflicts in the docs for a V1PodSpec) - but I guess that with 253 characters we can probably still use a uuid and save ourselves the trouble :)

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,43 @@
...
+    # overrides the executor setting.
+    retries = field(type=int,
+                    factory=int,
+                    mandatory=False,
+                    invariant=lambda r: (r >= 0, 'retries >= 0'))
+    volumes = field(type=PVector)
+
+    @property
+    def pod_name(self) -> str:
+        return self.name
```
keymone, line 25: previously this was `task_id`, which consisted of `name` and `uuid`, that task_id was then sent to mesos to uniquely identify a task. in kubernetes the corresponding entity is a pod. you will probably want to keep the same approach with kubernetes pods, having `pod_name` consist of a meaningful part and a unique id because pod_name must be unique in a namespace.

strictly speaking, `uuid` attribute doesn't need to be of UUID type, it's just that in mesos, but any random sequence of ascii will do if we want something more compact.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def set_pod_name(self, pod_name: str):
```
keymone, line 27: looking at this now, i don't think we use set_task_id anywhere, in mesos we create a task with a name and uuid is generated randomly. task_id value is then derived from those two. when state is recovered - it's `name` and `uuid` that are deserialized, not `task_id`.

you can probably just remove this method until we actually find a reason to have it.

task_processing/plugins/kubernetes/task_config.py
keymone, line 33: validation on composite fields like this belong in `__invariant__` method, see here: https://github.com/Yelp/task_processing/blob/cf843658d2af1122e7aeb279200a75aa60c785f5/task_processing/plugins/mesos/task_config.py#L48

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,43 @@
...
+            raise ValueError(f'Invalid format for pod_name {pod_name}')
+
+        if len(pod_name) > 253:
+            print(
+                f'Invalid format for pod_name {pod_name}.',
+                f'Pod_name must have up to 253 characters only.')
+            return False
+
+        regex_passed = re.match('[a-z0-9]([.-a-z0-9]*[a-z0-9])?', pod_name)
+        if not regex_passed:
```
keymone, line 40: validations on single field belong in per-field invariants: https://github.com/Yelp/task_processing/blob/cf843658d2af1122e7aeb279200a75aa60c785f5/task_processing/plugins/mesos/task_config.py#L92

keymone (changes requested):
re-reviewing :)

task_processing/plugins/kubernetes/task_config.py
keymone: as i recall, k8s itself won't do anything about pod name itself, it'll just error out if you try creating two pods with the same name in the same namespace.

uuids are probably an overkill, i would probably go for a random 6-byte base58 string these days.

we could try to be smart and recognize that those pods will be unique if you use retry number in pod_name, but then retries should be baked into task executor, while right now we have a pretty abstract implementation (even if it's in mesos folder): https://github.com/Yelp/task_processing/blob/cf843658d2af1122e7aeb279200a75aa60c785f5/task_processing/plugins/mesos/retrying_executor.py

nemacysts: oh, I was thinking of `generate_name` from https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ObjectMeta.md, but that won't work for us - a random 6-byte base58 string sounds good to me too, I was originally worried about the uuid length :p

task_processing/plugins/kubernetes/task_config.py
nemacysts: we use this in tron (https://github.com/Yelp/Tron/blob/5fd0b08ef8c7fd965960b105622c04499406aeb8/tron/mesos.py#L481) - iiuc, this is for recreating tron state after a restart

keymone: you're right, though if name and uuid are serialized in recovery data, we might still not need it, but it's not a blocker of course, i'm ok with keeping it

task_processing/plugins/kubernetes/task_config.py
lanceleih: I was looking for the paasta way of randomly generating the uuid in baseball 58, but could only find encoding in base64. So I just added it manually as there are no python libs for randomly generating base58 strings.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Aah thank you! Moving validation to __invariant__.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Done.

task_processing/plugins/kubernetes/task_config.py
lanceleih: Changed into invariant.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,43 @@
...
+        except ValueError:
+            raise ValueError(f'Invalid format for pod_name {pod_name}')
+
+        if len(pod_name) > 253:
+            print(
+                f'Invalid format for pod_name {pod_name}.',
+                f'Pod_name must have up to 253 characters only.')
+            return False
+
+        regex_passed = re.match('[a-z0-9]([.-a-z0-9]*[a-z0-9])?', pod_name)
```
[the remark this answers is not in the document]
lanceleih: Made as a constant.

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,57 @@
...
+                )
+            ),
+            (
+                re.match(REGEX_MATCH, conf.pod_name),
+                (
+                    f'Invalid format for pod_name {conf.pod_name}. ',
+                    f'Must comply with Kubernetes pod naming standards.'
+                )
+            )
+        )
```
nemacysts, line 33: super-nit: 
```suggestion
    def __invariant__(conf):
        return (
            (
                len(conf.pod_name) > MAX_POD_NAME_LENGTH,
                (
                    f'Invalid format for pod_name {conf.pod_name}. ',
                    f'Pod name must have up to {MAX_POD_NAME_LENGTH}.'
                )
            ),
            (
                re.match(REGEX_MATCH, conf.pod_name),
                (
                    f'Invalid format for pod_name {conf.pod_name}. ',
                    f'Must comply with Kubernetes pod naming standards.'
                )
            )
        )
        
```

nemacysts: we should probably add `black` to this repo to autoformat code at some point, I guess :p

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,57 @@
...
+    volumes = field(type=PVector)
+
+    @property
+    def pod_name(self) -> str:
+        return f'{self.name}.{self.uuid}'
+
+    def set_pod_name(self, pod_name: str):
+        try:
+            name, = pod_name.rsplit('.', maxsplit=1)
+            uuid = ''.join(secrets.choice(BASE_58) for i in range(UUID_BYTES))
```
nemacysts, line 53: we don't want to generate a new unique suffix here since callers of this function are already going to provide one for us so we want something like 
```suggestion
            name, uuid = pod_name.rsplit('.', maxsplit=1)
```

nemacysts: after that, the code that generates the "uuid" (i.e.,`''.join(secrets.choice(BASE_58) for i in range(UUID_BYTES))`) should probably be a function that we then use in the uuid field

e.g.:
```
def _generate_pod_suffix() -> str:
    return ''.join(secrets.choice(POD_SUFFIX_ALPHABET) for i in range(POD_SUFFIX_LENGTH))
    

class KubernetesTaskConfig(DefaultTaskConfigInterface):
    uuid = field(type=str, initial=_generate_pod_suffix)
    ...
```

since we only want task-proc to generate a "uuid" when creating a brand new task, not when editing an existing one

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,57 @@
+import re
+import secrets
+
+from pyrsistent import field
+from pyrsistent import PMap
+from pyrsistent import PVector
+
+from task_processing.interfaces.task_executor import DefaultTaskConfigInterface
+
+BASE_58 = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
```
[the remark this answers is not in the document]
lanceleih: Ah that's true. I missed that. Done!

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -0,0 +1,57 @@
...
+UUID_BYTES = 6
+MAX_POD_NAME_LENGTH = 253
+REGEX_MATCH = '[a-z0-9]([.-a-z0-9]*[a-z0-9])?'
```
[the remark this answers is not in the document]
lanceleih: Done.

tests/unit/plugins/kubernetes/kubernetes_task_config_test.py
```diff
@@ -0,0 +1,22 @@
...
+    except InvariantException as e:
+        print(e)
+        assert True
+
+    try:
+        k2 = k2.set(name="INVALID"+k2.name)
+        assert False, 'Must comply with Kubernetes pod naming standards.'
+    except InvariantException as e:
+        print(e)
+        assert True
```
nemacysts, line 22: ++ on writing tests!

suggestion: we probably want to split this into two tests just so that it's easier to see exactly what's broken should we make a bad refactor somewhere at some point. additionally, pytest has a nice wrapper for testing exceptions (https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions) that we might want to use. maybe something like:

```suggestion
def test_kubernetes_task_config_set_pod_name_rejects_long_name():
    task_config = KubernetesTaskConfig(name="fake_pod_name")

    with pytest.raises(InvariantException):
        task_config.set(name='a' * 254)

def test_kubernetes_task_config_enforces_kubernetes_name_requirements(     
    task_config = KubernetesTaskConfig(name="fake_pod_name")

    with pytest.raises(InvariantException):
        k2.set(name=f"INVALID{task_config.name})
```

nemacysts: (we'll need to add an `import pytest` line to the top of this file, of course)

nemacysts (approved):
it's probably also worth adding a test that ensures that the "happy path" works (i.e., `task_config.set_pod_name(pod_name="mock_pod.mock_uuid")` does the right thing), but this looks good as a starting point :)

tests/unit/plugins/kubernetes/kubernetes_task_config_test.py
lanceleih: Done.

nemacysts (approved):

PR #149, opened 2021-05-27
Remove TravisCI badge from README

nemacysts:
We no longer use Travis, so we might as well get rid of the badge :)


lanceleih (approved):
Ah true! Thanks for catching that.

PR #150, opened 2021-06-03
Kubernetes Pod Executor First Draft

lanceleih:
### Description
Implementing a Kubernetes Pod Executor to create a Kubernetes version of taskproc runner.


nemacysts:
oh, we'll also want to make the `__init__.py` I'm adding in https://github.com/Yelp/task_processing/pull/152 have a register_plugin method as in https://github.com/Yelp/task_processing/blob/master/task_processing/plugins/mesos/__init__.py#L11

nemacysts (approved):

PR #151, opened 2021-06-07
Enable GH Actions for all PRs + upload to PyPI

nemacysts:
Our previous config made it so that only PRs made from a branch on this
repo would get tested - so anyone creating a PR from their fork would
not get their changes tested automatically.

Additionally, add back in the config to upload to PyPI since Tron will 
need to pick up these builds for GH Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,14 @@
 name: CI
 
 # Controls when the action will run.
-on: [push, release]
+on:
+  push:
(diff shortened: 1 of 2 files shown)


danielpops (approved):
I've added the `pypi_password` secret to this repo https://fluffy.yelpcorp.com/i/CLRs3rvHvQJDr6xDpQ4GqfkfzGZzLQ15.png

PR #153, opened 2021-06-22
Add a KubeClient class

nemacysts:
Much like in paasta-tools, we want to ensure that anything talking to
the k8s API has loaded the correct kubeconfig so that it knows what
master to talk to as well as what user/certs to use.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,5 +35,6 @@ setup(
         'mesos_executor': ['addict', 'pymesos>=0.2.14', 'requests'],
         'metrics': ['yelp-meteorite'],
         'persistence': ['boto3'],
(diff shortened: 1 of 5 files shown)


nemacysts:
@lanceleih tI'm thinking that we do something similar to paasta-tools with how we used the k8s clientlib and wrap it all with this class that ensures that we've loaded a kubeconfig file

that said, lemme actually add this as a member in KubernetesPodExecutor

nemacysts:
ok, my thinking now is that when we initially implement `run()` we just do something like `self.kube_client.core.create_namespaced_pod(...)`

lanceleih (approved):
LGTM

PR #154, opened 2021-06-23
Add a way to track task metadata in k8s plugin

nemacysts:
We'll use this in the same way as we handled things in the mesos plugin:
run/kill will direfctly mutate the task_metadata member I'm adding in this PR
and we'll eventually have the thread(s?) that handle watching Pods also
add to this member in addition to the event queue that we'll share with
Tron

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -2,11 +2,15 @@ import threading
 import time
 from enum import auto
 from enum import unique
+from typing import Tuple
 
 from pyrsistent import field
 from pyrsistent import pmap
 from pyrsistent import PRecord
+from pyrsistent import PVector
+from pyrsistent import pvector
 from pyrsistent.typing import PMap
+from pyrsistent.typing import PVector as PVectorType
 
 from task_processing.interfaces import TaskExecutor
 from task_processing.plugins.kubernetes.kube_client import KubeClient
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -1,20 +1,65 @@
...
+    # what box this task/Pod was scheduled onto
+    node_name: str = field(type=str, initial='')
+
+    # the config used to launch this task/Pod
+    task_config: KubernetesTaskConfig = field(type=KubernetesTaskConfig, mandatory=True)
+
+    # TODO(TASKPROC-241): add current task state and task state history as we did for mesos
+    task_state: KubernetesTaskState = field(type=KubernetesTaskState, mandatory=True)
+    # Map of state to when that state was entered (stored as a timestamp)
+    task_state_history: PMap[str, int] = field(factory=pmap, mandatory=True)
```
keymone, line 31: it's like this in mesos, but i think using map was a mistake, list of (timestamp, state) would preserve history correctly in case when task bounces between same states. might be good chance to fix this.

keymone (approved):
lgtm, minor nit

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: I'd actually had that thought but assumed that we had done it that way on purpose in the mesos plugin!

fixed :)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -1,20 +1,72 @@
...
+    def run(self, task_config: KubernetesTaskConfig) -> None:
+        # we need to lock here since there will be other threads updating this metadata in response
+        # to k8s events
+        with self._lock:
+            self.task_metadata = self.task_metadata.set(
+                task_config.pod_name,
+                KubernetesTaskMetadata(
+                    task_config=task_config,
+                    task_state=KubernetesTaskState.TASK_PENDING,
+                    task_state_history=pvector(
```
keymone, line 55: you actually want to use `v` function here, as `pvector` will turn any sequence type (such as tuple) into PVector rather than forming a PVector with that sequence as first element:

    >>> from pyrsistent import pvector, v
    >>> t = (1,2,)
    >>> pvector(t)
    pvector([1, 2])
    >>> v(t)
    pvector([(1, 2)])

keymone (changes requested):

lanceleih (approved):
Thanks for specifying the args type! Everything else looks good to me 👍

keymone (approved):

PR #155, opened 2021-06-24
Teach k8s plugin how to kill a Pod

nemacysts:
This also adds in `namespace` as a required parameter for creating a k8s
executor instance.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -1,3 +1,4 @@
+import logging
 import threading
 import time
 from enum import auto
@@ -18,6 +19,8 @@ from task_processing.interfaces import TaskExecutor
 from task_processing.plugins.kubernetes.kube_client import KubeClient
 from task_processing.plugins.kubernetes.task_config import KubernetesTaskConfig
 from task_processing.utils import AutoEnum
+
+logger = logging.getLogger(__name__)
 
 
 @unique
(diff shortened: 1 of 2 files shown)


nemacysts:
we probably don't want to merge this until we have run() working so that we can actually run this code against a real cluster (at least until we have a working k8s itest setup in this repo)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -67,7 +70,22 @@ def reconcile(self, task_config: KubernetesTaskConfig) -> None:
...
     def kill(self, task_id: str) -> bool:
-        pass
+        """
+        Delete a Pod by name.
+
+        This function will have task_processing stop tracking the killed task/Pod and
+        will return True if the Pod was successfully killed (and False otherwise)
+        """
+        with self._lock:
+            self.task_metadata = self.task_metadata.discard(task_id)
```
keymone, line 80: we should probably only discard after we send request successfully and receive an event that confirms the deletion

nemacysts: ah, I was trying to figure out the exact flow of how the mesos plugin handled this - that makes sense tho

nemacysts: I guess I should also make sure that this delete call doesn't block too!

nemacysts: I need to do some testing, but I think that the return value of a `delete_*` call is just if the k8s api successfully processed the request - not if the underlying object was deleted. 

so the only change that we'll want to make here is deferring the metadata deletion to when we have a Watch setup to get Pod events if I'm understanding thing correctly!

keymone (approved):
lgtm, i'd probably add a log line where we attempt to delete the pod for completeness, but not a blocker

PR #158, opened 2021-06-30
Add skeleton of Pod event processing

nemacysts:
We'll have two threads: one that simply tails a k8s stream of Pod events
and places these into an internal queue and a second one that takes
these events and (potentially) converts them into task_processing Events
and/or takes some form of action on them (e.g., removing a Pod from the
plugin's task_metadata data structure)

For now the thread that creates Events is mostly just a stub though.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -1,61 +1,103 @@
 import logging
 import threading
 import time
-from enum import auto
-from enum import unique
-from typing import Tuple
+from queue import Queue
+from typing import Optional
+from typing import Union
 
+from kubernetes import watch
 from kubernetes.client import V1Status
-from pyrsistent import field
 from pyrsistent import pmap
-from pyrsistent import PRecord
-from pyrsistent import PVector
-from pyrsistent import pvector
 from pyrsistent import v
 from pyrsistent.typing import PMap
-from pyrsistent.typing import PVector as PVectorType
 
 from task_processing.interfaces import TaskExecutor
+from task_processing.interfaces.event import Event
 from task_processing.plugins.kubernetes.kube_client import KubeClient
 from task_processing.plugins.kubernetes.task_config import KubernetesTaskConfig
-from task_processing.utils import AutoEnum
+from task_processing.plugins.kubernetes.types import KubernetesTaskMetadata
+from task_processing.plugins.kubernetes.types import KubernetesTaskState
+from task_processing.plugins.kubernetes.types import PodEvent
+from task_processing.plugins.kubernetes.types import Sentinel
 
 logger = logging.getLogger(__name__)
-
-
-@unique
-class KubernetesTaskState(AutoEnum):
-    TASK_PENDING = auto()
-
-
-class KubernetesTaskMetadata(PRecord):
-    # what box this task/Pod was scheduled onto
-    node_name: str = field(type=str, initial='')
-
-    # the config used to launch this task/Pod
-    task_config: KubernetesTaskConfig = field(type=KubernetesTaskConfig, mandatory=True)
-
-    # TODO(TASKPROC-241): add current task state and task state history as we did for mesos
-    task_state: KubernetesTaskState = field(type=KubernetesTaskState, mandatory=True)
-    # Map of state to when that state was entered (stored as a timestamp)
-    task_state_history: PVectorType[Tuple[KubernetesTaskState, int]] = field(
-        type=PVector, factory=pvector, mandatory=True)
 
 
 class KubernetesPodExecutor(TaskExecutor):
     TASK_CONFIG_INTERFACE = KubernetesTaskConfig
 
-    def __init__(self, namespace: str) -> None:
-        self.kube_client = KubeClient()
+    def __init__(self, namespace: str, kube_config_path: Optional[str] = None) -> None:
+        self.kube_client = KubeClient(kube_config_path=kube_config_path)
         self.namespace = namespace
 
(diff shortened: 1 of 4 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -1,61 +1,103 @@
...
+        self.pending_event_processing_thread = threading.Thread(
+            target=self._pending_event_processing_loop,
+        )
+        self.pending_event_processing_thread.start()
+
+    def _pod_event_watch_loop(self) -> None:
+        # TODO(TASKPROC-243): we'll need to correctly handle resourceVersion expiration for the case
+        # where the gap between task_proc shutting down and coming back up is long enough for data
+        # to have expired from etcd
+        for pod_event in self.watch.stream(
```
keymone, line 68: wrap this in `while not self.stopping` and try/catch, otherwise we might be stuck with this thread dead and the rest of the system waiting for events that will never arrive.

keymone (approved):
lgtm

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: I wrapped both with `while not self.stopping` - we'll need to do some more work with this specific thread later to make sure that we try to resume from the correct resourceVersion - but we can do that when we start building in support for task_proc/tron restarts

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -1,61 +1,122 @@
...
+        while not self.stopping:
+            try:
+                for pod_event in self.watch.stream(
+                    self.kube_client.core.list_namespaced_pod,
+                    self.namespace
+                ):
+                    self.pending_events.put(pod_event)
+            except Exception:
+                logger.exception(
+                    "Exception encountered while watching Pod events - restarting watch!")
```
keymone, line 80: log the exception itself

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -1,61 +1,122 @@
...
+            logger.warning("Got unknown event with type: {event['type']}")
+
+        # TODO(TASKPROC-245): actually create real events to send to Tron by placing them in
+        # this plugin's event queue
+
+    def _pending_event_processing_loop(self) -> None:
+        event = None
+        while not self.stopping or not self.pending_events.empty():
+            try:
+                event = self.pending_events.get()  # blocks forever until an event is available
```
keymone, line 100: we need to wait with timeout, otherwise it'll never check if `self.stopping` became true

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -1,61 +1,122 @@
...
+                if event is Sentinel.POISON_PILL:
+                    return
+                # we need to narrow the types here since mypy isn't smart enough to see that the
+                # `is` check above means that we can no longer have anything but a PodEvent after it
+                assert not isinstance(event, Sentinel)
+
+                self._process_pod_event(event)
+            except Exception:
+                logger.exception(
+                    f"Exception encountered proccessing Pod events - skipping event {event} and "
```
keymone, line 111: log the exception details

keymone (changes requested):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: https://docs.python.org/3/library/logging.html#logging.Logger.exception will automatically log the traceback in addition to this message :)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: my train of thought was that this would be fine since if we're blocking here then there's no pending events and we can let the `join()` timeout in `stop()` and then just shutdown afterwards without fear of losing any data

if we should still add a timeout to this `get()`: we're doing a `join()` with a 1s timeout in `stop()`, so should we use the same timeout here or halve it (i.e, `get(timeout=.5)`)?

keymone: i see, ok. the problem is that .join may timeout not only because this thread is blocked on .get but also because it takes a long time to process an event. in that case timeout on .join will kill the thread in the middle of something which can lead to corruption.

or are you talking about self.pending_tasks.join? that will actually wait for queue to drop to zero. but then we need to be careful and coordinate shutdown such that no more events are placed on the queue after .join succeeds as this thread will immediately start working on new event and get killed in the middle of processing.

i guess i'd be a bit defensive here and just block on waiting + sleep in between while iterations so that we react to self.stopping reliably.

nemacysts: can do!

nemacysts: hmm, I guess this approach means that we no longer need to use a poison pill to ensure that we kill this thread since we'll constantly be checking if we should be stopping - i'm gonna go ahead and delete that code then

nemacysts: I actually ended up using `pending_events.join()` since that blocks until `task_done()` has been called sufficient times - not until the queue is empty

lemme know your thoughts on https://github.com/Yelp/task_processing/pull/158/commits/cd53f63bad75846b181d455cebccffbf0ff6af76!

lanceleih (approved):
lgtm

keymone (approved):

PR #159, opened 2021-07-05
Implement run in KubernetesPodExecutor

lanceleih:
### Description
Allows the Executor to launch pods with the task's image and command.

### Tests
`make test` - PASSED


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -48,11 +54,12 @@ class KubernetesPodExecutor(TaskExecutor):
     def __init__(self, namespace: str) -> None:
         self.kube_client = KubeClient()
         self.namespace = namespace
-
+        ensure_namespace(self.kube_client, namespace)
```
keymone, line 57: do we actually want this? i'd rather have this machinery fail if namespace is missing as that would indicate critical misconfiguration on another level.

nemacysts: we were just copying this pattern from paasta - but I guess maybe we should just create the namespace in puppet and assume that that's been applied anywhere where we're running tron?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -66,8 +73,26 @@ def run(self, task_config: KubernetesTaskConfig) -> None:
...
+            metadata=V1ObjectMeta(
+                name=task_config.pod_name,
+                namespace=self.namespace
+            ),
+            spec=V1PodSpec(
+                restart_policy=task_config.restart_policy,
+                containers=[container]
+            ),
+        )
+        self.api.create_namespaced_pod(namespace=self.namespace, body=pod)
```
nemacysts, line 92: we'll want to use `kube_client` here instead: 
```suggestion
        self.kube_client.core.create_namespaced_pod(namespace=self.namespace, body=pod)
```
so that we can ensure that there's no code path that tries to to k8s api things without loading a kubeconfig first (since creating a KubeClient will automatically do that for us)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -48,11 +54,12 @@ class KubernetesPodExecutor(TaskExecutor):
     def __init__(self, namespace: str) -> None:
         self.kube_client = KubeClient()
         self.namespace = namespace
-
+        ensure_namespace(self.kube_client, namespace)
         self.task_metadata: PMap[str, KubernetesTaskMetadata] = pmap()
+        self.api = CoreV1Api()
```
nemacysts, line 59: we'll want to use the instance of CoreV1Api from `self.kube_client` instead :)
```suggestion
```

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: I think we'll also want to wrap this in a `try-except`, log that the `run()` failed, and return `None` if this failed so that we don't crash Tron should the `run()` call error out in anyway (alternatively, we could probably return True/False if we were able to submit a request to k8s to run our pod since we know the pod name ahead of time anyway :p

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -39,3 +49,32 @@ def test_run_updates_task_metadata(k8s_executor):
...
+            namespace="task_processing_tests"
+        ),
+        spec=V1PodSpec(
+            restart_policy=task_config.restart_policy,
+            containers=[fake_container]
+        ),
+    )
+
+    assert k8s_executor.run(task_config) == task_config.pod_name
+    assert k8s_executor.api.create_namespaced_pod.call_args_list == [
```
nemacysts, line 78: once we've moved to using kube_client in run, this line will probably look something like:
```suggestion
    assert k8s_executor.kube_client.core.create_namespaced_pod.call_args_list
```

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -21,12 +25,18 @@ def k8s_executor():
         autospec=True
     ), mock.patch.dict(os.environ, {"KUBECONFIG": "/this/doesnt/exist.conf"}):
         executor = KubernetesPodExecutor(namespace="task_processing_tests")
+        executor.api.create_namespaced_pod = mock.Mock()
```
nemacysts, line 28: we probably won't need an extra mock here once we've moved to using kube_client
```suggestion
```

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
lanceleih: Removed 👍

task_processing/plugins/kubernetes/utils.py
```diff
@@ -0,0 +1,21 @@
+import logging
```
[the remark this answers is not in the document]
lanceleih: File deleted 👍

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -66,8 +73,26 @@ def run(self, task_config: KubernetesTaskConfig) -> None:
                     ),
                 ),
             )
-
-        # TODO(TASKPROC-231): actually launch a Pod
+        # TODO Add volume_devices and volume_mounts
```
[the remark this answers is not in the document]
lanceleih: Added this action to TASKPROC-238. Will add cpu, mem, and disk here too.

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
lanceleih: switched to using kube_client

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
lanceleih: switched to using kube_client

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
lanceleih: Added try-except that returns None if error occurs. Also shows the exception we run into.

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -66,8 +69,32 @@ def run(self, task_config: KubernetesTaskConfig) -> None:
...
+            spec=V1PodSpec(
+                restart_policy=task_config.restart_policy,
+                containers=[container]
+            ),
+        )
+
+        try:
+            self.kube_client.core.create_namespaced_pod(namespace=self.namespace, body=pod)
+        except Exception as e:
+            logger.error(f"Failed to create pod {task_config.pod_name}: {e}")
```
nemacysts, line 92: super-nit: we should probably be consistent and use `logger.exception` here since that's what we're using elsewhere (that'll also automatically include the traceback below the log message
```suggestion
        except Exception:
            logger.exception(f"Failed to create pod {task_config.pod_name}")
```

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -48,11 +52,10 @@ class KubernetesPodExecutor(TaskExecutor):
     def __init__(self, namespace: str) -> None:
         self.kube_client = KubeClient()
         self.namespace = namespace
-
         self.task_metadata: PMap[str, KubernetesTaskMetadata] = pmap()
         self._lock = threading.RLock()
 
-    def run(self, task_config: KubernetesTaskConfig) -> None:
+    def run(self, task_config: KubernetesTaskConfig) -> str:
```
nemacysts, line 58: we probably want to update the return type if we're gonna return the task name on success and None otherwise
```suggestion
    def run(self, task_config: KubernetesTaskConfig) -> Optional[str]:
```

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -37,8 +37,9 @@ def __invariant__(conf):
 
     uuid = field(type=str, initial=_generate_pod_suffix)  # type: ignore
     name = field(type=str, initial="default")
+    image = field(type=str)
```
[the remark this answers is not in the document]
lanceleih: Made mandatory 👍

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -42,6 +53,51 @@ def test_run_updates_task_metadata(k8s_executor):
...
+        uuid="fake_id",
+        image="fake_docker_image",
+        command="fake_command"
+    )
+    k8s_executor.kube_client.core.create_namespaced_pod.side_effect = ApiException(
+        status=403, reason="Fake unauthorized message")
+    assert k8s_executor.run(task_config) is None
+    assert mock_logger.exception.call_args(mock.call(
+        f"Failed to create pod {task_config.pod_name}:",
+        f"{k8s_executor.kube_client.core.create_namespaced_pod.side_effect}"))
```
nemacysts, line 98: its probably fine to not test that we logged something - this test will break if we ever change the wording even tho no functionality has changed
```suggestion
```

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -2,6 +2,12 @@
 from unittest import mock
 
 import pytest
+from kubernetes.client import ApiException
+from kubernetes.client import V1Container
+from kubernetes.client import V1ObjectMeta
+from kubernetes.client import V1Pod
+from kubernetes.client import V1PodSpec
+from mock import patch
```
nemacysts, line 10:
```suggestion
```

we probably want to use `mock` from the stdlib (e.g., what we've imported with `from unittest import mock`

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -42,6 +53,51 @@ def test_run_updates_task_metadata(k8s_executor):
...
+        ),
+    )
+
+    assert k8s_executor.run(task_config) == task_config.pod_name
+    assert k8s_executor.kube_client.core.create_namespaced_pod.call_args_list == [
+        mock.call(body=fake_pod, namespace='task_processing_tests')
+    ]
+
+
+@patch("task_processing.plugins.kubernetes.kubernetes_pod_executor.logger", autospec=True)
```
nemacysts, line 85: we can probably get rid of this if we no longer check that we logged something

nemacysts (commented):
looking good!

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -153,8 +155,32 @@ def run(self, task_config: KubernetesTaskConfig) -> None:
                     ),
                 ),
             )
+        # TODO (TASKPROC-238): Add volume, cpu, gpu, desk, mem, etc.
+        container = V1Container(
+            image=task_config.image,
+            name=task_config.name,
+            command=[task_config.command]
```
nemacysts, line 162: I was messing around with this branch so that i could more easily test something else and realized that we'll need to do one of two things: 
* use [shlex.split](https://docs.python.org/3.6/library/shlex.html#shlex.split) to get a list
* or, change KubernetesTaskConfig to have `command` be a PVector and have Tron pass us a list of `["cmd", "any", "desired", "args"]` (i.e., have Tron/our caller to the shlex.split())

otherwise, if we wanted to run something like `sleep 120` as a command, right now this code would try to run a command called `sleep 120` which won't exist:
```
    State:          Terminated
      Reason:       ContainerCannotRun
      Message:      OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"sleep 120\": executable file not found in $PATH": unknown```

Comment (keymone) [reply task_processing/plugins/kubernetes/kubernetes_pod_executor.py:162]: i think the closer to configuration step we do this, the better. best would be setting command to list[str] in paasta when generating tron configs.

Comment (nemacysts) [reply task_processing/plugins/kubernetes/kubernetes_pod_executor.py]: hmm, so looking closer at what people are doing in soaconfigs, it looks like there's several occurrences of constructs like `cmd && cmd2` - it turns out that in mesos we were relying on the fact that the containerizer would wrap the command in `/bin/sh -c` (https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L647)

I think we'll want to do something similar - have command be `["/bin/sh", "-c"]` and then we pass in the command from the KubernetesTaskConfig into `args` e.g.:
```Python
container = V1Container(
            image=task_config.image,
            name=task_config.name,
            # for compatibility with Mesos, we wrap all commands in a shell so that users can give us things like
            # cmd1 && cmd2
            command=["/bin/sh", "-c"],
            args=[task_config.command],
)
```

nemacysts: (bonus of doing this change: we don't need to change anything elsewhere - although we probably want to have `setup_tron_namespace`/tron eventually pass us `/bin/sh -c`  as a command explicitly at some point in the future)

nemacysts (approved):

PR #162, opened 2021-07-12
Handle certificate expiry in k8s plugin

nemacysts:
We regualrly rotate certificates in our clusters and the k8s clientlib
won't notice when these files change (nor does it have a way to add a
hook on a 401), so manually catch any 401 API exceptions and reload the
kubeconfig file.

### DIFF
--- a/task_processing/plugins/kubernetes/kube_client.py
+++ b/task_processing/plugins/kubernetes/kube_client.py
@@ -1,8 +1,17 @@
+import logging
 import os
+from http import HTTPStatus
 from typing import Optional
 
 from kubernetes import client as kube_client
 from kubernetes import config as kube_config
+from kubernetes.client.exceptions import ApiException
+from kubernetes.client.models.v1_pod import V1Pod
+from kubernetes.client.models.v1_status import V1Status
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_ATTEMPTS = 2
 
 
 class KubeClient:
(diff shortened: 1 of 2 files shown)


nemacysts:
I have a second commit for handling cert expiry in run, but that's slightly blocked on #159

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -173,25 +191,44 @@ def kill(self, task_id: str) -> bool:
...
+                    logger.info(
+                        "Recieved UNAUTHORIZED response from apiserver - assuming certs have "
+                        "expired and reloading."
+                    )
+                    self.kube_client.reload_kubeconfig()
+                else:
+                    logger.exception(
+                        f"Failed to request termination for Pod {task_id} due to unhandled API "
+                        "exception, retrying."
+                    )
```
keymone, line 224: a matter of taste, but this being repeated few times, looks like it could be factored out into a function

keymone (approved):
minor nit, otherwise lgtm

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: I moved things over to the KubeClient class so that there's less logic in the plugin code itself, but I wasn't able to get rid of 100% of the duplication - but at least now it's just the general structure that's duplicated and not the exact same code

task_processing/plugins/kubernetes/kube_client.py
```diff
@@ -14,12 +23,133 @@ def __init__(self, kubeconfig_path: Optional[str] = None) -> None:
...
+        Currently, the only exception that will cause a reload is an ApiException with an
+        HTTP 401 (unauthorized) status.
+        """
+        if isinstance(exception, ApiException):
+            if exception.status == HTTPStatus.UNAUTHORIZED.value:
+                logger.info(
+                    "Recieved UNAUTHORIZED response from apiserver - assuming certs have "
+                    "expired and reloading."
+                )
+                self.reload_kubeconfig()
```
lanceleih, line 69: Maybe we can have some kind of confirmation maybe that reloading the kubeconfig file worked successfully? Load_kube_config doesn't return anything, but what happens if load_kube_config file doesn't find the file for example. Not sure if this is overkill though.

nemacysts: in that case, the k8s clientlib would throw an exception - i'm not quite sure what we want to do in that case though - maybe once we have event stuff working we can catch that exception and send a control event in that case and tell tron to shut us down?

lanceleih: Yup. I think that sounds reasonable 👍

lanceleih (approved):

jfongatyelp (approved):

PR #163, opened 2021-07-12
Add additional task states to KubernetesTaskState

nemacysts:
Much like in the Mesos plugin, we're going to try to track the different
states that a task can be in - and to do this, we need to expand our
enum for the potential set of states for a k8s task

### DIFF
--- a/task_processing/plugins/kubernetes/types.py
+++ b/task_processing/plugins/kubernetes/types.py
@@ -29,6 +29,10 @@ class PodEvent(TypedDict):
 @unique
 class KubernetesTaskState(AutoEnum):
     TASK_PENDING = auto()
+    TASK_RUNNING = auto()
+    TASK_FINISHED = auto()
+    TASK_FAILED = auto()
+    TASK_UNKNOWN = auto()
 
 
 class KubernetesTaskMetadata(PRecord):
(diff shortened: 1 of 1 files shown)


task_processing/plugins/kubernetes/types.py
```diff
@@ -38,8 +42,7 @@ class KubernetesTaskMetadata(PRecord):
     # the config used to launch this task/Pod
     task_config: KubernetesTaskConfig = field(type=KubernetesTaskConfig, mandatory=True)
 
-    # TODO(TASKPROC-241): add current task state and task state history as we did for mesos
     task_state: KubernetesTaskState = field(type=KubernetesTaskState, mandatory=True)
-    # Map of state to when that state was entered (stored as a timestamp)
-    task_state_history: PVectorType[Tuple[KubernetesTaskState, int]] = field(
+    # List of state to when that state was entered (stored as a timestamp)
+    task_state_history: PVectorType[Tuple[KubernetesTaskState, float]] = field(
```
keymone, line 47: float to preserve sub-second resolution?

also now that task state history is a list, we should probably think about limits on number of elements

keymone (approved):

task_processing/plugins/kubernetes/types.py
nemacysts: it’s a float to match the timestamp in the Event class we bubble up to Tron - the Pyrsistent field for that is marked as a float, so it complains that an int isn’t a valid type. I thought it might be a little confusing to do `time.time()` in some places and `int(time.time()` in others, so I decided to be consistent and not cast to an int

this probably belongs in the PR that adds in emitting Event objects to the event queue, but that’s slightly blocked on some other work so i haven’t published it yet :p

nemacysts:
whoops, I merged this into the wrong branch

PR #164, opened 2021-07-14
Start translating k8s events into task_processing ones

nemacysts:
We'll use these in Tron in order to transition a task between the
various states in Tron's state machine

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -29,6 +29,12 @@ logger = logging.getLogger(__name__)
 POD_WATCH_THREAD_JOIN_TIMEOUT_S = 1.0
 POD_EVENT_THREAD_JOIN_TIMEOUT_S = 1.0
 QUEUE_GET_TIMEOUT_S = 0.5
+SUPPORTED_POD_MODIFIED_EVENT_PHASES = {
+    "Failed",
+    "Running",
+    "Succeeded",
+    "Unknown",
+}
 
 
 class KubernetesPodExecutor(TaskExecutor):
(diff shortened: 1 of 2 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -107,21 +108,175 @@ def _pod_event_watch_loop(self) -> None:
...
+                platform_type="killed"
+            )
+        )
+
+    def __handle_modified_pod_event(self, event: PodEvent) -> None:
+        pod = event["object"]
+        pod_name = pod.metadata.name
+        task_metadata = self.task_metadata[pod_name]
+
+        if pod.status.phase not in {"Succeeded", "Failed", "Running", "Unknown"}:
```
lanceleih, line 136: minor-nit: Can this be declared in a variable?
`POD_PHASE = {"Succeeded", "Failed", "Running", "Unknown"}`

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -107,21 +114,175 @@ def _pod_event_watch_loop(self) -> None:
...
+                    task_config=task_metadata.task_config,
+                    platform_type="running"
+                )
+            )
+            return
+
+        # XXX: figure out how to handle this correctly (and when this actually
+        # happens - we were unable to cajole k8s into giving us an event with an Unknown
+        # phase)
+        elif pod.status.phase == "Unknown":
```
jfongatyelp, line 219: Should we be checking if we'd already set the task_metadata to Unknown as well, just in case?

jfongatyelp (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: yea, won't hurt!

jfongatyelp (approved):

PR #165, opened 2021-07-16
Add resource limits, caps, env vars, and volumes to k8s task_config

nemacysts:
We'll need these for:
* resource limits: to isolate resource usage per-task as we did in Mesos
* capabilties: to match the security status-quo in PaaSTA/Tron-on-Mesos
* volumes: to allow people to mount files on disk :p
* environment variables: we'll need these for both PaaSTA secrets as
  well as for normal usage (some tasks may configure
  themselves/libraries with environment variables)

### DIFF
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -165,7 +165,7 @@ class KubernetesTaskConfig(DefaultTaskConfigInterface):
         type=float,
         initial=128.0,
         factory=float,
-        invariant=lambda m: (m >= 128, 'mem is >= 128'))
+        invariant=lambda m: (m >= 32, 'mem is >= 32'))
     disk = field(
         type=float,
         initial=10.0,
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/task_config.py
```diff
@@ -37,21 +131,64 @@ def __invariant__(conf):
...
+    cpus = field(
+        type=float,
+        initial=0.1,
+        factory=float,
+        invariant=lambda c: (c > 0, 'cpus > 0'))
+    memory = field(
+        type=float,
+        initial=128.0,
+        factory=float,
+        invariant=lambda m: (m >= 128, 'mem is >= 128'))
```
jfongatyelp, line 168: Oh is this a kubernetes limitation, 128 is the minimum? We do have some jobs configured for less right now it seems.

jfongatyelp (approved):

task_processing/plugins/kubernetes/task_config.py
nemacysts: oh, I was just arbitrarily picking something "reasonable" since the Mesos plugin default seemed unreasonably low - but then i forgot to check what our existing values look like to see if the Mesos defaults still make sense :p

jfongatyelp (approved):

PR #168, opened 2021-07-21
TASKPROC-249: Add support to pull secrets from kubernetes in KubernetesPodExecutor

jfongatyelp:
~(*Note to reviewers*: This branch is based off of work to be merged in #166 as it depends on some of the utilities defined there)~ this PR is now based off master 

This adds support for setting envvar contents from kubernetes secrets.  This requires an external sync that creates the secrets in our expected namespace with the correct format, which will be handled outside of task_processing.

This PR assumes that the external sync will sync the secrets into Kubernetes Secret resources, with naming convention: `{k8s namespace}-secret-{secret_grouping}-{secret_name}`, where `secret_grouping` is typically at the service level (or Tron namespace level), or under the `_shared` global grouping. 

Our general error handling if we cannot create a pod will catch the situation where we are attempting to create a pod with an environment variable referencing a nonexistent secret, so no additional error handling was added in this PR.

### DIFF
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -191,6 +191,11 @@ class KubernetesTaskConfig(DefaultTaskConfigInterface):
     )
 
     @property
+    def prefix(self) -> str:
+        # For tron jobs, this is the tron namespace
+        return f'{self.name.split(".")[0]}'
+
+    @property
     def pod_name(self) -> str:
         return f'{self.name}.{self.uuid}'  # type: ignore
 
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -191,11 +191,6 @@ class KubernetesTaskConfig(DefaultTaskConfigInterface):
     )
 
     @property
-    def prefix(self) -> str:
-        # For tron jobs, this is the tron namespace
-        return f'{self.name.split(".")[0]}'
-
-    @property
     def pod_name(self) -> str:
         return f'{self.name}.{self.uuid}'  # type: ignore
 
--- a/task_processing/plugins/kubernetes/utils.py
+++ b/task_processing/plugins/kubernetes/utils.py
@@ -64,9 +64,9 @@ def is_shared_secret(value: str) -> bool:
 
 
 def get_secret_kubernetes_env_var(
-    key: str, value: str, task_prefix: str, namespace: str,
+    key: str, value: str, task_name: str, namespace: str,
 ) -> V1EnvVar:
-    task_prefix = task_prefix if not is_shared_secret(value) else SHARED_SECRET_SERVICE
+    task_prefix = task_name.split('.')[0] if not is_shared_secret(value) else SHARED_SECRET_SERVICE
     sanitised_task_prefix = get_sanitised_kubernetes_name(task_prefix)
 
     secret = get_secret_name_from_ref(value)
@@ -85,7 +85,7 @@ def get_secret_kubernetes_env_var(
 
 
 def get_kubernetes_env_vars(
-    environment: PMap[str, str], task_prefix: str, namespace: str,
+    environment: PMap[str, str], task_name: str, namespace: str,
 ) -> List[V1EnvVar]:
     """
     Given a dict of environment variables, transform them into the corresponding Kubernetes
@@ -102,7 +102,7 @@ def get_kubernetes_env_vars(
     # to be able to find them in k8s for any readers/users outside of Yelp
     secret_env_vars = [
         get_secret_kubernetes_env_var(
-            key=key, value=value, task_prefix=task_prefix, namespace=namespace
+            key=key, value=value, task_name=task_name, namespace=namespace
         )
         for key, value
         in environment.items()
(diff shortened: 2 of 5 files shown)


task_processing/plugins/kubernetes/utils.py
```diff
@@ -63,7 +66,48 @@ def is_secret_env_var(value: str) -> bool:
...
+    task_prefix = get_sanitised_kubernetes_name(task_prefix)
+    secret = get_secret_name_from_ref(value)
+    sanitised_secret = get_sanitised_kubernetes_name(secret)
+    if is_shared_secret(value):
+        task_prefix = get_sanitised_kubernetes_name(SHARED_SECRET_SERVICE)
+        return V1EnvVar(
+            name=key,
+            value_from=V1EnvVarSource(
+                secret_key_ref=V1SecretKeySelector(
+                    name=f"{namespace}-secret-{task_prefix}-{sanitised_secret}",
```
keymone, line 89: is this needed?

keymone (approved):
lgtm

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -190,6 +190,11 @@ def __invariant__(conf):
         invariant=_valid_capabilities,
     )
 
+    @property
+    def prefix(self) -> str:
```
nemacysts, line 194: are there any other places we could use this prefix outside of secrets? if not, maybe we should do a more specific name? e.g., `secret_X`?

task_processing/plugins/kubernetes/utils.py
nemacysts: something like: 
```python
    task_prefix = task_prefix if not is_shared_secret(value) else SHARED_SECRET_SERVICE
    sanitised_task_prefix = get_sanitised_kubernetes_name(task_prefix)

    secret = get_secret_name_from_ref(value)
    sanitised_secret = get_sanitised_kubernetes_name(secret)

    return V1EnvVar(
        name=key,
        value_from=V1EnvVarSource(
            secret_key_ref=V1SecretKeySelector(
                name=f"{namespace}-secret-{sanitised_task_prefix}-{sanitised_secret}",
                key=secret,
                optional=False,
            )
        ),
    )
```

would let us get rid of the duplication here :)

task_processing/plugins/kubernetes/task_config.py
jfongatyelp: This is a good call, because we are really only ever going to use this for secrets, I've just skipped this altogether and passed the full task_config.name to our secret utils instead.

task_processing/plugins/kubernetes/utils.py
keymone: i was actually asking specifically about `name` arg, docs say it's optional

keymone (approved):

task_processing/plugins/kubernetes/utils.py
jfongatyelp: This is mostly just copied from [paasta](https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L909-L947).

I believe this name is the name of the actual V1Secret resource we want to get the value_from, since `key` is strictly the name defined within the secret data of that V1Secret resource, so we do need to keep it here (I wonder what a `SecretKeySelector` would pull from without a name).

Names of the secrets will be determined by https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py

task_processing/plugins/kubernetes/utils.py
```diff
@@ -52,7 +55,38 @@ def is_secret_env_var(value: str) -> bool:
     return SECRET_VALUE_REGEX.match(value) is not None
 
 
-def get_kubernetes_env_vars(environment: PMap[str, str]) -> List[V1EnvVar]:
+def get_secret_name_from_ref(value: str) -> str:
+    return value.split("(")[1][:-1]
+
+
+def is_shared_secret(value: str) -> bool:
```
nemacysts, line 62: maybe we should add a note here for the benefit of non-Yelpers looking at this code about what shared secrets are?
e.g.,
"""
In some cases, multiple services need access to the same secrets - rather than storing these N times internally, Yelp has a mechanism in which a secret can be tagged as "shared" and thus only need to be modified in one place should that secret need to be rotated/updated/etc.

These "shared" secrets are referenced using the format `SHARED_SECRET(secret_name)` instead of the more tightly-scoped format of `SECRET(secret_name)`
"""

nemacysts (approved):

task_processing/plugins/kubernetes/utils.py
keymone: ok, in that case i think this is problematic. task_processing being a library should make minimal amount of assumptions as to how entities in kubernetes (or anywhere else) are named, dependencies like that should be inverted, users of task_processing client should pass in either a function that will construct the correct name from relevant inputs or pass in a mapping where the name can be looked up.

the way this code works means that if we change how paasta generates configs (absolutely reasonable thing to do, as those are entities created and managed by paasta) will break users of task_processing. coordinating rollout of such changes will become a nightmare.

cc @jfongatyelp @nemacysts

keymone: a followup thought - why do we even have to deal with paasta's secrets system in task_processing?

(1) ysoa-configs -> (2) paasta pre-processing -> (3) tronfig -> (4) task-processing

i think secret references should be prepared in step (2) just like we replace service/instance with correct docker image and stuff

PR #169, opened 2021-07-21
Ensure that k8s pod names follow yelp conventions

nemacysts:
We now implicitly call get_sanitised_kubernetes_name() when returning
pod names so that we don't need our users to do this.

This change does mean that `name != pod_name.split(".")[0]`, but this
should be fine

### DIFF
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -13,6 +13,8 @@ from pyrsistent import pmap
 from pyrsistent import PVector
 from pyrsistent import pvector
 from pyrsistent import v
+
+from task_processing.plugins.kubernetes.utils import get_sanitised_kubernetes_name
 if TYPE_CHECKING:
     from task_processing.plugins.kubernetes.types import DockerVolume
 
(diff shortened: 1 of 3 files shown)


jfongatyelp (approved):

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -113,16 +115,16 @@ def _valid_capabilities(capabilities: Sequence[str]) -> Tuple[bool, Optional[str
...
             (
-                len(conf.pod_name) < MAX_POD_NAME_LENGTH,
+                len(get_sanitised_kubernetes_name(self.pod_name)) < MAX_POD_NAME_LENGTH,
                 (
                     f'Pod name must have up to {MAX_POD_NAME_LENGTH} characters.'
                 )
             ),
             (
-                re.match(VALID_POD_NAME_REGEX, conf.pod_name),
+                re.match(VALID_POD_NAME_REGEX, get_sanitised_kubernetes_name(self.pod_name)),
```
jfongatyelp, line 127: Sorry I didn't notice this earlier; should we be double-sanitising here, since `pod_name` is already `get_sanitised_kubernetes_name(self.name)`?

PR #170, opened 2021-07-22
TASKPROC-249: Revert duplication of paasta logic, pass secret refs directly

jfongatyelp:
in #168 , we duplicated some paasta implementation within task-proc to set up secrets in envvars.  

This wasn't going to be maintainable moving forward, so we've removed that and task_proc will simply be passed a new `secrets_environment` config specifying the name of the envvar and the required `V1SecretKeySelector` values necessary to pull from a `V1Secret`.

This makes task_processing code much cleaner and straightforward.

### DIFF
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -108,7 +108,7 @@ def _valid_volumes(volumes: Sequence["DockerVolume"]) -> Tuple[bool, Optional[st
 
 
 def _valid_secret_envs(secret_envs: Mapping[str, "SecretEnvSource"]) -> Tuple[bool, Optional[str]]:
-    # XXX: Note we are not validating existence of secret in k8s here, leave that to creation of pod
+    # Note we are not validating existence of secret in k8s here, leave that to creation of pod
     for key, value in secret_envs.items():
         if set(value.keys()) != VALID_SECRET_ENV_KEYS:
             return (
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -1,6 +1,7 @@
 import re
 import secrets
 import string
+from typing import Mapping
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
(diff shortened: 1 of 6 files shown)


task_processing/plugins/kubernetes/task_config.py
```diff
@@ -104,6 +107,18 @@ def _valid_volumes(volumes: Sequence["DockerVolume"]) -> Tuple[bool, Optional[st
     return (True, None)
 
 
+def _valid_secret_envs(secret_envs: Mapping[str, "SecretEnvSource"]) -> Tuple[bool, Optional[str]]:
+    # XXX: Note we are not validating existence of secret in k8s here, leave that to creation of pod
```
nemacysts, line 111: unless we want to want to hit the k8s api here, probably safe to remove the `XXX` TODO marker :)
```suggestion
    # Note we are not validating existence of secret in k8s here, leave that to creation of pod
```

task_processing/plugins/kubernetes/utils.py
```diff
@@ -126,17 +57,21 @@ def get_kubernetes_env_vars(
...
-        get_secret_kubernetes_env_var(
-            key=key, value=value, task_name=task_name, namespace=namespace
+        V1EnvVar(name=key, value_from=V1EnvVarSource(
+            secret_key_ref=V1SecretKeySelector(
+                name=value["secret"],
+                key=value["key"],
+                optional=False,
+            ),
+        ),
         )
```
nemacysts, line 70: super-nit: can we update the formatting a tiny bit here? we should probably add `black` to this repo at some point tbh though :p
```suggestion
        V1EnvVar(
            name=key,
            value_from=V1EnvVarSource(
                secret_key_ref=V1SecretKeySelector(
                    name=value["secret"],
                    key=value["key"],
                    optional=False,
                ),
            ),
        )
```

nemacysts (approved):
nice :D

task_processing/plugins/kubernetes/utils.py
jfongatyelp: Ah, right.  I do think *something* in our precommits likes to change this out from under me, because this annoys me as well (why would closing params for nested calls end at the same indentation?!).  I'll try to see if I can work around it.

nemacysts: worst case, we can merge this as-is and just get around to running black on this repo next week and stop caring about this :p

jfongatyelp: Turns out this one was my fault, since `secret_key_ref` was not indented correctly to begin with.  precommits did not argue argue with me this time!

nemacysts (approved):

PR #175, opened 2021-08-04
TASKPROC-244:  Initialize metadata via state data and implement reconcile for k8s

jfongatyelp:
In order to handle tron restarts, we will need to first pass kubernetes_pod_executor a list of KubernetesTaskConfigs from our saved state.

We initialize this in `kubernetes_pod_executor` upon init, filling `task_metadata` with UNKNOWN KubernetesTaskStates for each pod name to speed up initalization.

Reconciliation with said pods will then happen in `reconcile()`, where we will check the state of the pod we are attempting to reconcile with the Kubernetes API, then re-use the exact same logic to set TaskState (etc) as when we handle `PodEvents` based on pod phases.   This required splitting out said logic into two new helpers, `__update_deleted_pod` and `__update_modified_pod` so they could be used for both usecases.

### DIFF
--- a/task_processing/plugins/kubernetes/kube_client.py
+++ b/task_processing/plugins/kubernetes/kube_client.py
@@ -14,6 +14,10 @@ logger = logging.getLogger(__name__)
 DEFAULT_ATTEMPTS = 2
 
 
+class ExceededMaxAttempts(Exception):
+    pass
+
+
 class KubeClient:
     def __init__(self, kubeconfig_path: Optional[str] = None) -> None:
         kubeconfig_path = kubeconfig_path or os.environ.get("KUBECONFIG")
@@ -157,6 +161,7 @@ class KubeClient:
     def get_pod(
         self, namespace: str, pod_name: str, attempts: int = DEFAULT_ATTEMPTS,
     ) -> Optional[V1Pod]:
+        max_attempts = attempts
         while attempts:
             try:
                 pod = self.core.read_namespaced_pod(
@@ -180,6 +185,6 @@ class KubeClient:
                 logger.exception(
                     f"Failed to fetch pod {pod_name} due to unhandled exception."
                 )
-                return None
+                raise
         logger.info(f"Ran out of retries attempting to fetch pod {pod_name}.")
-        return None
+        raise ExceededMaxAttempts(f'Retried fetching pod {pod_name} {max_attempts} times.')
--- a/task_processing/plugins/kubernetes/kube_client.py
+++ b/task_processing/plugins/kubernetes/kube_client.py
@@ -168,8 +168,6 @@ class KubeClient:
                     namespace=namespace, name={pod_name},
                 )
                 if not pod:
-                    # XXX: We do not currently distinguish between not finding a pod and hitting api
-                    #      exception when fetching pod
                     logger.info(f"Found no pods matching {pod_name}.")
                     return None
                 else:
--- a/task_processing/plugins/kubernetes/kube_client.py
+++ b/task_processing/plugins/kubernetes/kube_client.py
@@ -167,12 +167,12 @@ class KubeClient:
                 pod = self.core.read_namespaced_pod(
                     namespace=namespace, name={pod_name},
                 )
-                if not pod:
+                return pod
+            except ApiException as e:
+                # Unknown pod throws ApiException w/ 404
+                if e.status == 404:
                     logger.info(f"Found no pods matching {pod_name}.")
                     return None
-                else:
-                    return pod
-            except ApiException as e:
                 if not self.maybe_reload_on_exception(exception=e) and attempts:
                     logger.debug(
                         f"Failed to fetch pod {pod_name} due to unhandled API exception, retrying",
--- a/task_processing/plugins/kubernetes/kube_client.py
+++ b/task_processing/plugins/kubernetes/kube_client.py
@@ -153,3 +153,33 @@ class KubeClient:
 
         logger.info(f"Ran out of retries attempting to create {pod.metadata.name}.")
         return False
+
+    def get_pod(
+        self, namespace: str, pod_name: str, attempts: int = DEFAULT_ATTEMPTS,
+    ) -> Optional[V1Pod]:
+        while attempts:
+            try:
+                pod = self.core.read_namespaced_pod(
+                    namespace=namespace, name={pod_name},
+                )
+                if not pod:
+                    # XXX: We do not currently distinguish between not finding a pod and hitting api
+                    #      exception when fetching pod
+                    logger.info(f"Found no pods matching {pod_name}.")
+                    return None
+                else:
+                    return pod
+            except ApiException as e:
+                if not self.maybe_reload_on_exception(exception=e) and attempts:
+                    logger.debug(
+                        f"Failed to fetch pod {pod_name} due to unhandled API exception, retrying",
+                        exc_info=True
+                    )
+                attempts -= 1
+            except Exception:
+                logger.exception(
+                    f"Failed to fetch pod {pod_name} due to unhandled exception."
+                )
+                return None
+        logger.info(f"Ran out of retries attempting to fetch pod {pod_name}.")
+        return None
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -434,7 +434,8 @@ class KubernetesPodExecutor(TaskExecutor):
         try:
             pod = self.kube_client.get_pod(namespace=self.namespace, pod_name=pod_name)
         except Exception:
-            logger.exception(f"Cannot reconcile pod {pod_name}")
+            logger.exception(f"Hit an exception attempting to fetch pod {pod_name}")
+            pod = None
 
         if pod_name not in self.task_metadata:
             self._initialize_existing_task(task_config)
@@ -448,7 +449,6 @@ class KubernetesPodExecutor(TaskExecutor):
                 )
             )
 
-        with self.task_metadata_lock:
             if not pod:
                 # Pod has gone away while restarting
                 logger.info(
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -233,7 +233,7 @@ class KubernetesPodExecutor(TaskExecutor):
             self.task_metadata = self.task_metadata.set(
                 pod_name,
                 task_metadata.set(
-                    node_name=pod.status.host_ip,
+                    node_name=pod.spec.node_name,
                     task_state=KubernetesTaskState.TASK_RUNNING,
                     task_state_history=task_metadata.task_state_history.append(
                         (KubernetesTaskState.TASK_RUNNING, time.time()),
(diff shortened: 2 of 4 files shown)


nemacysts:
@jfongatyelp 
> 1. If we cannot find the pod after restart, then we treat it as a `killed` pod (i.e. manual operator intervention.).  Should this instead be treated as `lost`?

yea, that's what I was thinking (treating it as `lost`) - this *should* match what we do in Mesos with `unknown` (we just took over `unknown` for a different purpose in k8s :p)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -133,17 +165,23 @@ def __handle_deleted_pod_event(self, event: PodEvent) -> None:
...
     def __handle_modified_pod_event(self, event: PodEvent) -> None:
         pod = event["object"]
+        self.__update_modified_pod(pod=pod, event=event)
+
+    def __update_modified_pod(self, pod: V1Pod, event: Optional[PodEvent]) -> None:
+        """ Called during reconciliation and normal event handling """
         pod_name = pod.metadata.name
         task_metadata = self.task_metadata[pod_name]
 
+        raw_event = event['raw_object'] if event else ''  # TODO: when is this empty?
```
jfongatyelp, line 183: Whoops, forgot to remove this TODO but I did have a question -- do you think it's worth crafting an event string here that indicates this came from reconciliation? 
I could not find where we use this in tron on the kubernetes side, but on the mesos side we strictly used it as a string in error reporting.

nemacysts: `raw_object` is normally just a V1Pod in `dict` form at the moment for `task` events and probably not very useful for most people :p 

I think that if we want to report on reconciliation events, we'd maybe want to send a `control` events, but probably only if we want to take action on the fact that we reconciled/didn't reconcile something

imo, maybe we just want to send `None` in `raw` for reconciled events? that way, the data structures are a little more clear - Optional[Dict] instead of Union[Dict, str]

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -394,7 +432,24 @@ def reconcile(self, task_config: KubernetesTaskConfig) -> None:
...
+            pod_name,
+            task_metadata.set(
+                task_config=task_config
+            )
+        )
+
+        if not pod:
+            # Pod has gone away while restarting
+            logger.info(f"Pod {pod_name} for task {task_config.name} was no longer found.")
+            self.__update_deleted_pod(pod_name=pod_name, event=None)
```
nemacysts: I think we'll want to add a helper for this case and send a `lost` event instead of treating this like a deleted pod event

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -266,3 +281,75 @@ def test_process_event_enqueues_task_processing_events_deleted(
...
+    assert all(
+        [
+            tm.task_state == KubernetesTaskState.TASK_UNKNOWN
+            for tm in k8s_executor_with_tasks.task_metadata.values()
+        ]
+    )
+
+
+def test_reconcile_running_pod(
+    k8s_executor_with_tasks
```
nemacysts, line 324: thoughts on having the fixture yield a tuple of (executor, task_configs)?

task_processing/plugins/kubernetes/kube_client.py
```diff
@@ -153,3 +153,33 @@ def create_pod(
...
+        self, namespace: str, pod_name: str, attempts: int = DEFAULT_ATTEMPTS,
+    ) -> Optional[V1Pod]:
+        while attempts:
+            try:
+                pod = self.core.read_namespaced_pod(
+                    namespace=namespace, field_selector=f'metadata.name={pod_name}'
+                ).items
+                if not pod:
+                    # XXX: We do not currently distinguish between not finding a pod and hitting api
+                    #      exception when fetching pod
```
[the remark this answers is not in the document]
jfongatyelp: Yeah. I added a log line when we explicitly dont' find it so at least there's that, but otherwise downstream code won't know the difference -- is that a big issue?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -45,12 +46,37 @@
...
+                for pod_name in tasks:
+                    logger.debug(
+                        f"Initializing blank task metadata for known pod {pod_name}"
+                    )
+                    # We are initiating with blank config and state on initial load, then leave
+                    # matching tasks to running/completed pods in reconcile()
+                    self.task_metadata = self.task_metadata.set(
+                        pod_name,
+                        KubernetesTaskMetadata(
+                            task_config=KubernetesTaskConfig(image='', command='noop command'),
```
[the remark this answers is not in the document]
jfongatyelp: Yeah this felt weird doing this, I like being able to tell this was explicitly due to reconciling and not some other issue setting image/command.

task_processing/plugins/kubernetes/kube_client.py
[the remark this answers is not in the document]
nemacysts: ah - it's probably worth re-raising the ApiException instead of always returning None

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
jfongatyelp: Cool, I have just left these `None` for reconciled events.

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
jfongatyelp: This worked nicely, and it let me test all 4 pod phases we handle much easier.

task_processing/plugins/kubernetes/kube_client.py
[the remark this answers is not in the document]
jfongatyelp: Alright I've gone this route though I wasn't clear what we wanted to do on the reconcile() side -- should we try to just remove the pod from task_metadata or leave it be at that point in `TASK_UNKNOWN` forever? 

Will we be trying to reconcile all known running pods on some interval?

jfongatyelp:
Lots of updates just applied --
* Additional exception handling; we now throw exceptions when we cannot find a pod via k8s API instead of just treating it like a pod that cannot be found
* Tests updated as suggested in review comments
* Fixed multiple instances of not reassigning pmap variables to results of set() (set() does not modify variables in place)
* Now setting KubernetesTaskMetadata.node_name from pod.spec.node_name instead of pod.status.host_ip
* Fixed `get_pod` implementation (leftovers from a previous use of list_namespaced_pods instead of read_namespaced_pod)

jfongatyelp:
New updates per our discussion of what we're going to restore from state -- no longer just pod names, but instead taking in full task_configs.

lanceleih (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -391,10 +430,53 @@ def run(self, task_config: KubernetesTaskConfig) -> Optional[str]:
...
     def reconcile(self, task_config: KubernetesTaskConfig) -> None:
-        # TASKPROC(244): we probably only want reconcile() to fill in the task_config
-        # member of the relevant KubernetesTaskMetadata and do an unconditional state
-        # reconciliation on startup
-        pass
+        pod_name = task_config.pod_name
+        try:
+            pod = self.kube_client.get_pod(namespace=self.namespace, pod_name=pod_name)
+        except Exception:
+            logger.exception(f"Cannot reconcile pod {pod_name}")
```
nemacysts, line 437: should we set `pod = None` here so that we correctly enter line 452?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -391,10 +430,53 @@ def run(self, task_config: KubernetesTaskConfig) -> Optional[str]:
...
+                        terminal=False,
+                        timestamp=time.time(),
+                        raw=None,
+                        task_config=task_metadata.task_config,
+                        platform_type="lost"
+                    )
+                )
+            else:
+                # Treat like a modified pod
+                self.__update_modified_pod(pod=pod, event=None)
```
nemacysts, line 479: just curious: how come we're not just nesting this in the above context manager?

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -266,3 +301,73 @@ def test_process_event_enqueues_task_processing_events_deleted(
...
+
+def test_reconcile_existing_pods(
+    k8s_executor, mock_task_configs
+):
+
+    mock_pods = []
+    test_phases = ['Running', 'Succeeded', 'Failed', 'Unknown']
+    for i in range(len(mock_task_configs)):
+        taskconf = mock_task_configs[i]
+        phase = test_phases[i]
```
nemacysts, line 352: thoughts on something like
```suggestion
    for taskconf, phase in zip(mock_task_configs, test_phases):
```
explicitly indexing in python always looks a bit off to me, but feel free to drop this issue :)

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -266,3 +301,73 @@ def test_process_event_enqueues_task_processing_events_deleted(
...
+        k8s_executor,
+        "kube_client",
+        autospec=True
+    ) as mock_kube_client:
+        mock_kube_client.get_pod.side_effect = mock_pods
+        for taskconf in mock_task_configs:
+            k8s_executor.reconcile(taskconf)
+
+    assert k8s_executor.event_queue.qsize() == 4
+    assert len(k8s_executor.task_metadata) == 2
```
nemacysts, line 370: might be worth adding a comment here about what two tasks we're expecting to be in task_metadata just in case this ever fails in the future since this test has a decent bit of setup and its not immediately clear

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
jfongatyelp: This was originally just `return` ing here but lost the line during some updates, which would have left the action in a perpetual unknown state.  

I think it seems more user-friendly to treat this as LOST, so I'll go your `pod = None` route.

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
jfongatyelp: Ah, this was an error, there is no reason not to all work within the same context manager.

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -199,7 +238,7 @@ def __handle_modified_pod_event(self, event: PodEvent) -> None:
             self.task_metadata = self.task_metadata.set(
                 pod_name,
                 task_metadata.set(
-                    node_name=pod.status.host_ip,
+                    node_name=pod.spec.node_name,
```
nemacysts, line 241: nice, thank you for fixing this!

nemacysts (approved):

PR #177, opened 2021-08-16
Add node selectors and node affinities

kawaiwanyelp:
### Description
This PR adds node selectors and affinities to the k8s task config, and makes the pod executor use them. Couple of things:

- By default, it does not add any selectors or affinities. If we want to set, say, a default selector for the default pool, that should be set upstream.
- In soa-configs, node selectors and affinities are exposed to users as just node selectors. In task_processing, they are kept separate. The assumption is that the soa-configs interface has already been transpiled to selectors and affinities upstream.
- I moved the task metadata stuff to its own file because it was royally screwing with typing imports.

### Testing
`make test`

### Notes
no real-world test yet, but i am reasonably confident this works, since this is exactly how paasta does it. ill still do a test before merging, but the PR is in a reviewable state

### DIFF
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -135,17 +135,41 @@ def _valid_capabilities(capabilities: Sequence[str]) -> Tuple[bool, Optional[str
 def _valid_node_affinities(affinities: Sequence["NodeAffinity"]) -> Tuple[bool, Optional[str]]:
     for aff in affinities:
         missing_keys = REQUIRED_NODE_AFFINITY_KEYS.difference(set(aff.keys()))
-        if len(missing_keys) > 0:
+        if missing_keys:
             return (
                 False,
                 f"Invalid node affinity: got {aff} but missing keys {missing_keys}"
             )
-        if aff["operator"] not in NodeAffinityOperator.ALL:
+
+        op, val = aff["operator"], aff["value"]
+        if op not in NodeAffinityOperator:
+            valid_operators = list(o.value for o in NodeAffinityOperator)
             return (
                 False,
-                f"Invalid node affinity operator: got {aff['operator']}, "
-                f"but expected one of: {NodeAffinityOperator.ALL}",
+                f"Invalid node affinity operator: got '{op}', "
+                f"but expected one of: {valid_operators}",
             )
+
+        elif (
+            op in {NodeAffinityOperator.IN, NodeAffinityOperator.NOT_IN} and
+            type(val) != list
+        ):
+            return (
+                False,
+                "Invalid node affinity value: "
+                f"got non-list value '{val}' for affinity operator '{op}'"
+            )
+
+        elif (
+            op in {NodeAffinityOperator.GT, NodeAffinityOperator.LT} and
+            type(val) != int
+        ):
+            return (
+                False,
+                "Invalid node affinity value: "
+                f"got non-int value '{val}' for affinity operator '{op}'",
+            )
+
     return True, None
 
 
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -15,6 +15,8 @@ from pyrsistent import PVector
 from pyrsistent import pvector
 from pyrsistent import v
 
+from task_processing.plugins.kubernetes.types import NodeAffinity
+from task_processing.plugins.kubernetes.types import NodeAffinityOperator
 from task_processing.plugins.kubernetes.utils import get_sanitised_kubernetes_name
 if TYPE_CHECKING:
     from task_processing.plugins.kubernetes.types import DockerVolume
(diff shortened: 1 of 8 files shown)


task_processing/plugins/kubernetes/types.py
```diff
@@ -27,6 +16,22 @@ class SecretEnvSource(TypedDict):
...
+    DOES_NOT_EXIST = "DoesNotExist"
+    GT = "Gt"
+    LT = "Lt"
+    ALL = {IN, NOT_IN, EXISTS, DOES_NOT_EXIST, GT, LT}
+
+
+class NodeAffinity(TypedDict):
+    key: str
+    operator: str
+    value: Any
```
nemacysts, line 32: just curious: do the k8s docs mention what values are expected? although, i guess that even if they do we'd have a pretty ugly Union here instead of an Any

task_processing/plugins/kubernetes/utils.py
```diff
@@ -144,3 +150,49 @@ def get_pod_volumes(volumes: PVector['DockerVolume']) -> List[V1Volume]:
...
+                f"got invalid affinity operator '{op}'; "
+                f"expected one of: {NodeAffinityOperator.ALL}"
+            )
+
+        match_expressions.append(
+            V1NodeSelectorRequirement(key=str(aff["key"]), operator=op, values=val)
+        )
+
+    # package into V1NodeAffinity
+    if len(match_expressions) == 0:
```
nemacysts, line 190: super-nit-feel-free-to-ignore:
```suggestion
    if not match_expressions:
```

task_processing/plugins/kubernetes/utils.py
```diff
@@ -144,3 +150,49 @@ def get_pod_volumes(volumes: PVector['DockerVolume']) -> List[V1Volume]:
...
+        return None
+    term = V1NodeSelectorTerm(match_expressions=match_expressions)
+    selector = V1NodeSelector(node_selector_terms=[term])
```
nemacysts, line 193: we don't do anything else with `term`, so we could just inline it when creating the selector - but i don't feel very strongly about this

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -136,13 +144,18 @@ def test_run(k8s_executor):
...
         ),
     )
 
     assert k8s_executor.run(task_config) == task_config.pod_name
     assert k8s_executor.kube_client.core.create_namespaced_pod.call_args_list == [
         mock.call(body=expected_pod, namespace='task_processing_tests')
     ]
+    assert mock_get_node_affinity.call_args_list == [
+        mock.call(pvector([dict(key="a_label", operator="In", value=[])])),
+    ]
```
nemacysts, line 158: thoughts on just relying on the check that create_namespaced_pod was called with the expected body and not checking how get_node_affinity was called?

nemacysts (commented):
🔥

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -129,6 +132,23 @@ def _valid_capabilities(capabilities: Sequence[str]) -> Tuple[bool, Optional[str
     return (True, None)
 
 
+def _valid_node_affinities(affinities: Sequence["NodeAffinity"]) -> Tuple[bool, Optional[str]]:
+    for aff in affinities:
+        missing_keys = REQUIRED_NODE_AFFINITY_KEYS.difference(set(aff.keys()))
+        if len(missing_keys) > 0:
```
lanceleih, line 138:
```suggestion
        if len(missing_keys):
```

task_processing/plugins/kubernetes/utils.py
```diff
@@ -144,3 +150,49 @@ def get_pod_volumes(volumes: PVector['DockerVolume']) -> List[V1Volume]:
...
+    match_expressions = []
+    for aff in affinities:
+        op = aff["operator"]
+        val = aff["value"]
+
+        if op in {NodeAffinityOperator.IN, NodeAffinityOperator.NOT_IN}:
+            if type(val) != list:
+                raise TypeError(
+                    f"got non-list value '{val}' for affinity operator '{op}'"
+                )
```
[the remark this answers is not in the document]
lanceleih: +1

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
kawaiwanyelp: i generally prefer checking how things are called to validate correct args, in addition to checking output. open to changing

task_processing/plugins/kubernetes/types.py
kawaiwanyelp: [k8s docs](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1NodeSelectorRequirement.md) ask for list[str], with additional reqs for the diff operators.

lanceleih (approved):

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
nemacysts: sounds good! I normally only do one or the other since if you can assert the output you know everything is working as intended and you don't have extra test failures if you happened to change the function signature of something

i'm fine with keeping this as is tho!

task_processing/plugins/kubernetes/types.py
nemacysts: yea, leaving this as Any seems fine then

task_processing/plugins/kubernetes/types.py
```diff
@@ -27,6 +17,37 @@ class SecretEnvSource(TypedDict):
...
+    def __contains__(cls, v):
+        try:
+            cls(v)
+        except ValueError:
+            return False
+        else:
+            return True
+
+
+class NodeAffinityOperator(str, enum.Enum, metaclass=EnumSet):
```
nemacysts, line 30: oh nice - TIL!

nemacysts (approved):

PR #179, opened 2021-08-19
Allow node affinity values to be wrapped in tuples

nemacysts:
The way that Tron config works, lists often get turned into tuples. The
k8s clientlib doesn't seem to mind that it's getting tuples here, so we
might as well allow them to be passed in in the first place so that our
users (or us in Tron) don't need to care about tuples vs lists in the
first place here.


kawaiwanyelp (approved):
If the k8s API is fine, I'm fine with it.

PR #180, opened 2021-08-30
Hardcode the DNSPolicy for Pods to Default

nemacysts:
This is what is recommended by our internal security team, so let's just
hardcode for now since we *shouldn't* really need any of the other
policies.


lanceleih (approved):
LGTM. Just need to get the itests error resolved.

nemacysts:
itests are mesos-only and flaky, so I'm gonna go ahead and merge this

PR #181, opened 2021-08-30
Set additional security-related Pod settings

nemacysts:
These are exactly the same as what we do in PaaSTA.


nemacysts:
I also tested this with my local tron instance and everything was :gucci:

lanceleih (approved):

nemacysts:
(I just fixed a merge commit)

PR #183, opened 2021-09-03
Truncate long pod names

nemacysts:
We don't appear to be anywhere near this limit, but might as well
future-proof while we can

### DIFF
--- a/task_processing/plugins/kubernetes/utils.py
+++ b/task_processing/plugins/kubernetes/utils.py
@@ -81,18 +81,42 @@ def get_kubernetes_env_vars(
     return env_vars + secret_env_vars
 
 
-def get_sanitised_kubernetes_name(name: str, replace_dots: bool = False) -> str:
+def get_sanitised_kubernetes_name(
+    name: str,
+    replace_dots: bool = False,
+    replace_forward_slash: bool = False,
+    length_limit: int = 0,
+) -> str:
     """
     Helper to ensure that any names given to Kubernetes objects follow our conventions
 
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/utils.py
```diff
@@ -81,18 +81,42 @@ def get_kubernetes_env_vars(
...
+    NOTE: For names exceeding the length limit, we'll truncate the name and replace the
+    truncated portion with a unique suffix.
     """
     name = name.replace("_", "--")
     if name.startswith("--"):
         name = name.replace("--", "underscore-", 1)
     if replace_dots:
         name = name.replace(".", "dot-")
+    if replace_forward_slash:
+        name = name.replace("/", "slash-")
```
lanceleih, line 108: Can we change this from `slash-` to `-slash-` and `-dot-` to make it more human readable?

nemacysts: I was mostly keeping the same transforms that we do in paasta  where we do `volume_name.replace("/", "slash-").replace(".", "dot-")`

we're sorta limited on space (63 characters for docker volume names), so it's probably worth keeping these in their current "ugly" format - additionally, (imo) it's nice to be consistent with how these are named in paasta just for familiarities sake

that said, we can definitely add a `-` prefix, but the added length might cause us to truncate in some places (which is even less readable :p)

lanceleih: Ah I hadn't realized this was the same as in paasta. In that case, it makes sense to be more consistent 👍

lanceleih (approved):

PR #184, opened 2021-09-03
Emit running event if k8s compacts one away

nemacysts:
If a Pod exits quickly enough, kubernetes will compact the running +
finished events into a single finished event. This messes up the Tron
state machine as it assumes that if a `done` transition happens on a
pending task, then that task failed to start.

We can remediate this by ensuring that we send out a `running` event if
task_proc has not processed a running event by the time it sees a finish
one.


nemacysts:
I tested this on my local tron with jobs that just ran `ls` or `date` - and those jobs correctly transitioned through the state machine (and I saw the log message I added here)

keymone (approved):

PR #186, opened 2021-12-01
Add Service Account support

nemacysts:
We'll use this for Pod Identity - there's a webhook that will inject
a secret token + some environment variables if a Pod has a Service
Account set up.

### DIFF
--- a/task_processing/plugins/kubernetes/task_config.py
+++ b/task_processing/plugins/kubernetes/task_config.py
@@ -32,7 +32,8 @@ POD_SUFFIX_LENGTH = 6
 # or (max filename length) - (max namespace length) -(pod UID length) - (separators)
 # but let's give ourselves a little buffer so we'll round down a bit
 MAX_POD_NAME_LENGTH = 150
-VALID_POD_NAME_REGEX = '[a-z0-9]([.-a-z0-9]*[a-z0-9])?'
+MAX_DNS_SUBDOMAIN_NAME_LENGTH = 253
+VALID_DNS_SUBDOMAIN_NAME_REGEX = '[a-z0-9]([.-a-z0-9]*[a-z0-9])?'
 VALID_VOLUME_KEYS = {'mode', 'container_path', 'host_path'}
 VALID_SECRET_ENV_KEYS = {'secret_name', 'key'}
 VALID_CAPABILITIES = {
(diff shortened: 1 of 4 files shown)


task_processing/plugins/kubernetes/task_config.py
```diff
@@ -32,7 +32,8 @@
 # or (max filename length) - (max namespace length) -(pod UID length) - (separators)
 # but let's give ourselves a little buffer so we'll round down a bit
 MAX_POD_NAME_LENGTH = 150
-VALID_POD_NAME_REGEX = '[a-z0-9]([.-a-z0-9]*[a-z0-9])?'
+MAX_DNS_SUBDOMAIN_NAME_LENGTH = 253
+VALID_DNS_SUBDOMAIN_NAME_REGEX = '[a-z0-9]([.-a-z0-9]*[a-z0-9])?'
```
jfongatyelp, line 36: I think this might need ^ and $ to actually test the whole name, or we might want to use re.fullmatch?

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -116,6 +116,7 @@ def test_run(mock_get_node_affinity, k8s_executor):
         annotations={
             "paasta.yelp.com/some_annotation": "some_value",
         },
+        service_account_name="test_sa",
```
jfongatyelp, line 119: Technically this isn't valid with the `_` right? Which I think means our regex match is only matching the first few valid chars here since this wasn't throwing any issues

tests/unit/plugins/kubernetes/kubernetes_task_config_test.py
```diff
@@ -228,3 +228,36 @@ def test_valid_node_affinities_invalid_affinity(node_affinity, exc_msg):
             node_affinities=[node_affinity]
         )
     assert exc_msg in exc.value.invariant_errors[0]
+
+
+@pytest.mark.parametrize(
+    "service_account_name", (
+        "",
+        "F" * 300,
```
jfongatyelp, line 236: Should we also test one that doesnt' match the DNS regex?

danielpops (approved):
lgtm, no strong opinions

tests/unit/plugins/kubernetes/kubernetes_task_config_test.py
nemacysts: done :)

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
nemacysts: wow, can't believe this evaded us for so long

jfongatyelp (approved):

PR #187, opened 2022-01-21
Remove deadsnakes dependency, and fix make tox_integration

TheNavigat:


examples/cluster/playground/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial
+FROM docker-dev.yelpcorp.com/xenial_yelp
```
mks-m, line 1: as this is public repo, can we use ubuntu:bionic here? it should have python3.6 packages

mks-m (changes requested):

mks-m (approved):

PR #188, opened 2022-02-04
Add support for file-based secrets

nemacysts:
For PaaSTA services, we have support for mounting Kubernetes Secrets as
files directly - this adds similar support for anything launched by
task_processing.

For Tron usage, we're expecting paasta-tools to deal with formatting all
the information that we need here and pass it to Tron (which will then
just wrap it in a KubernetesTaskConfig)


danielpops (approved):
:ultrafastshipitparrot:

Molaire:
Tested with tron and paasta! LGTM

Molaire (approved):

PR #190, opened 2022-03-23
Enable exposing container ports

nemacysts:
This generally won't be needed for most of Yelp's task_processing
usecases - but it is needed for how we're planning on integrating Spark
with Tron.

Spark drivers expose several ports for normal operation, and we need to
launch containers with these ports exposed.

### DIFF
--- a/tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
+++ b/tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
@@ -6,6 +6,7 @@ from kubernetes.client import ApiException
 from kubernetes.client import V1Affinity
 from kubernetes.client import V1Capabilities
 from kubernetes.client import V1Container
+from kubernetes.client import V1ContainerPort
 from kubernetes.client import V1HostPathVolumeSource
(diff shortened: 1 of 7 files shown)


tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -15,6 +15,7 @@
 from kubernetes.client import V1SecurityContext
 from kubernetes.client import V1Volume
 from kubernetes.client import V1VolumeMount
+from kubernetes.client.models.v1_container_port import V1ContainerPort
```
jfongatyelp, line 18: Oh, weird, how come we had to import from `models` unlike all our other V1* objects (V1VolumeMount, V1SecurityContext, etc)

jfongatyelp (approved):

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
nemacysts: oh, my bad - this was my LSP importing the full path instead of the "nice" path...lemme fix this

PR #193, opened 2022-07-14
Implement empty dir support mount for k8s

anjaberens:
This is based off: https://github.com/Yelp/task_processing/pull/188


task_processing/plugins/kubernetes/task_config.py
```diff
@@ -118,6 +120,22 @@ def _valid_volumes(volumes: Sequence["DockerVolume"]) -> Tuple[bool, Optional[st
...
 
+def _valid_empty_volumes(volumes: Sequence[EmptyVolume]) -> Tuple[bool, Optional[str]]:
+    for volume in volumes:
+        if set(volume.keys()) != VALID_EMPTY_VOLUME_KEYS:
+            return (
+                False,
+                f'Invalid empty volume format, must only contain following keys: '
+                f'{VALID_EMPTY_VOLUME_KEYS}, got: {volume.keys()}'
+            )
+        if volume.get('medium') not in [None, "Memory"]:
```
nemacysts, line 131: might be nice to make the valid mediums into a set and extract it to a named constant (like `VALID_DOCKER_VOLUME_MODES`)

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -118,6 +120,22 @@ def _valid_volumes(volumes: Sequence["DockerVolume"]) -> Tuple[bool, Optional[st
...
+            return (
+                False,
+                "Invalid medium for empty volume, must be one of [None, 'Memory']",
```
nemacysts, line 134: and then use that named constant here

task_processing/plugins/kubernetes/types.py
```diff
@@ -12,6 +13,12 @@ class DockerVolume(TypedDict):
     mode: str  # XXX: Literal["RO", "RW"] once we drop older Python support
 
 
+class EmptyVolume(TypedDict):
+    container_path: str
+    medium: Optional[str]  # XXX: Optional[Literal["Memory"]] In this case
+    size: Optional[str]
```
nemacysts, line 19: maybe add a TODO for adding validation for this?

nemacysts (commented):
nice, looking good!

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -118,6 +121,27 @@ def _valid_volumes(volumes: Sequence["DockerVolume"]) -> Tuple[bool, Optional[st
...
 
+def _valid_empty_volumes(volumes: Sequence[EmptyVolume]) -> Tuple[bool, Optional[str]]:
+    for volume in volumes:
+        if not set(volume.keys()) <= VALID_EMPTY_VOLUME_KEYS:
+            return (
+                False,
+                f'Invalid empty volume format, must only contain following keys: '
+                f'{VALID_EMPTY_VOLUME_KEYS}, got: {volume.keys()}'
+            )
+        if 'container_path' not in set(volume.keys()):
```
nemacysts, line 132: oh, not all the keys are required? i think i'd prefer using `issubset` over the operator version if so - i think this is rare enough that most people (myself included) will need to look at the docs to remember what <= does in a set context

nemacysts, line 132: i think this can be 
```suggestion
        if 'container_path' not in volume:
```

`in` on a dict will automatically call .keys() and .keys provides a dict view (which is set-like) so we don't need to turn this into a set or call keys()

task_processing/plugins/kubernetes/utils.py
```diff
@@ -190,6 +192,49 @@ def get_pod_volumes(volumes: PVector['DockerVolume']) -> List[V1Volume]:
...
+    unique_volumes: Dict[str, 'EmptyVolume'] = {
+        get_sanitised_volume_name(f"empty--{volume['container_path']}", length_limit=63): volume
+        for volume in empty_volumes
+    }
+
+    return [
+        V1Volume(
+            name=name,
+            empty_dir=V1EmptyDirVolumeSource(
+                medium=volume.get('medium', ""),
```
nemacysts, line 230: shouldn't the default here be None?

anjaberens: The default is &quot;&quot; which means to use the node's default medium. Must be an empty string (default) or Memory.

https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EmptyDirVolumeSource.md

that's how i interpreted this one.

task_processing/plugins/kubernetes/task_config.py
anjaberens: I still want to prove that its strictly a subset of valid empty volume keys since if i only do 'container_path' in sneaky other keys can sneak in.

task_processing/plugins/kubernetes/utils.py
nemacysts: ah cool!

nemacysts (approved):

PR #194, opened 2022-07-19
Specify other containers to run

tzhu-yelp:
Allow the user to run more than one container in the pod for a task. Since the pod and container configuration is pretty much all flattened in KubernetesTaskConfig, I've made additional containers addable by nesting KubernetesTaskConfig itself. Of course, none of the pod-level configuration will ever get applied in a nested KubernetesTaskConfig, and nesting more isn't supported (actually checked against). 

Right now, common environment/volume configs would have to be duplicated at the caller. I am not sure if doing something like automatically applying such configs from the main task config would make much sense (I think it would be a bit surprising). 

Another note is that the main container is still called "main", but containers specified via extra_containers need to be given a name via the keys of extra_containers.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -32,6 +32,7 @@ from task_processing.plugins.kubernetes.utils import get_kubernetes_env_vars
 from task_processing.plugins.kubernetes.utils import get_kubernetes_volume_mounts
 from task_processing.plugins.kubernetes.utils import get_node_affinity
 from task_processing.plugins.kubernetes.utils import get_pod_volumes
+from task_processing.plugins.kubernetes.utils import get_sanitised_kubernetes_name
 from task_processing.plugins.kubernetes.utils import get_security_context_for_capabilities
 
 logger = logging.getLogger(__name__)
@@ -431,7 +432,10 @@ class KubernetesPodExecutor(TaskExecutor):
             containers = [self._create_container_definition("main", task_config)]
 
             for name, nested_config in task_config.extra_containers.items():
-                containers.append(self._create_container_definition(name, nested_config))
+                containers.append(self._create_container_definition(
+                    get_sanitised_kubernetes_name(name),
+                    nested_config,
+                ))
 
             pod = V1Pod(
                 metadata=V1ObjectMeta(
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -392,36 +392,47 @@ class KubernetesPodExecutor(TaskExecutor):
 
         logger.debug("Exiting Pod event processing - stop requested.")
 
+    def _create_container_definition(
+        self,
+        name: str,
+        task_config: KubernetesTaskConfig,
+    ) -> V1Container:
+        return V1Container(
+            image=task_config.image,
+            name=name,
+            command=["/bin/sh", "-c"],
+            args=[task_config.command],
+            security_context=get_security_context_for_capabilities(
+                cap_add=task_config.cap_add,
+                cap_drop=task_config.cap_drop,
+            ),
+            resources=V1ResourceRequirements(
+                limits={
+                    "cpu": task_config.cpus,
+                    "memory": f"{task_config.memory}Mi",
+                    "ephemeral-storage": f"{task_config.disk}Mi",
+                }
+            ),
+            env=get_kubernetes_env_vars(
+                environment=task_config.environment,
+                secret_environment=task_config.secret_environment,
+                field_selector_environment=task_config.field_selector_environment,
+            ),
+            volume_mounts=get_kubernetes_volume_mounts(task_config.volumes),
+            ports=[V1ContainerPort(container_port=port) for port in task_config.ports],
+        )
+
     def run(self, task_config: KubernetesTaskConfig) -> Optional[str]:
         try:
-            container = V1Container(
-                image=task_config.image,
-                # XXX: we were initially planning on using the name from KubernetesTaskConfig here,
-                # but its too easy to go over the length limit for container names (63 characters),
-                # so we're just hardcoding something for now since container names aren't used for
-                # anything at the moment
-                name="main",
-                command=["/bin/sh", "-c"],
-                args=[task_config.command],
-                security_context=get_security_context_for_capabilities(
-                    cap_add=task_config.cap_add,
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/task_config.py
```diff
@@ -254,6 +254,16 @@ def __invariant__(self) -> Tuple[Tuple[bool, str], ...]:
...
 
+    extra_containers = field(
+        type=PMap if not TYPE_CHECKING else PMap[str, "KubernetesTaskConfig"],
+        initial=m(),
+        factory=pmap,
+        invariant=lambda containers: (
+            not any([container.extra_containers for container in containers.values()]),
+            'extra_containers cannot have extra_containers',
+        ),
+    )
```
nemacysts, line 265: can we add a test for this?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -392,36 +392,47 @@ def _pending_event_processing_loop(self) -> None:
...
         logger.debug("Exiting Pod event processing - stop requested.")
 
+    def _create_container_definition(
+        self,
+        name: str,
+        task_config: KubernetesTaskConfig,
+    ) -> V1Container:
+        return V1Container(
+            image=task_config.image,
+            name=name,
```
nemacysts, line 402: i think we'll want to go ahead and sanitize this name (what i was trying to be lazy and avoid doing by naming the original container "main") :p 

container names need to be valid [RFC 1123](https://www.rfc-editor.org/rfc/rfc1123#section-2) DNS labels - https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/vendor/sigs.k8s.io/kustomize/kyaml/yaml/internal/k8sgen/pkg/util/validation/validation.go?L190 is what k8s does right now (tl;dr: 63 char max and is captured by this regex: `[a-z0-9]([-a-z0-9]*[a-z0-9])?`

we should be able to use [get_sanitised_kubernetes_name](https://sourcegraph.yelpcorp.com/Yelp/task_processing@8ab2a98/-/blob/task_processing/plugins/kubernetes/utils.py#L101:5) to do this tho

nemacysts: that said - i'm not particularly sure what i'd prefer: that we truncate (i.e., use `get_sanitised_kubernetes_name` or that we reject too-long names at validation. 

i think i'd lean towards truncation? it's what we do pretty much everywhere else, so seems fine to continue doing that here

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -392,36 +392,47 @@ def _pending_event_processing_loop(self) -> None:
...
+                    "ephemeral-storage": f"{task_config.disk}Mi",
+                }
+            ),
+            env=get_kubernetes_env_vars(
+                environment=task_config.environment,
+                secret_environment=task_config.secret_environment,
+                field_selector_environment=task_config.field_selector_environment,
+            ),
+            volume_mounts=get_kubernetes_volume_mounts(task_config.volumes),
+            ports=[V1ContainerPort(container_port=port) for port in task_config.ports],
```
nemacysts, line 422: hmm, do we need to validate against multiple containers trying to expose the same port?

nemacysts (commented):
this is a pretty clever approach! 

i'm a little worried about KubernetesTaskConfig being somewhat multi-purpose now (main container/scheduling pod config + optional sidecar config) - but given how often we needed to extend task_proc in the past for mesos (read: not really), this is probably an unfounded worry since i don't think we'll really end up designing ourselves into a hole here

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nattofriends: Oh, is that not allowed

tzhu-yelp: Oh, is that something that is not allowed?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
tzhu-yelp: I think if people aren't actually going to refer to the names exactly, then it should be fine?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: i'm...not sure - i can take a quick look, was mostly just thinking out loud

tests/unit/plugins/kubernetes/kubernetes_task_config_test.py
```diff
@@ -30,7 +30,7 @@ def test_kubernetes_task_config_set_pod_name_truncates_long_name():
     assert task_config.pod_name != "a" * 1000
 
 
-def test_kubernetes_task_config_enforces_command_requirmenets():
+def test_kubernetes_task_config_enforces_command_requirments():
```
nemacysts, line 33:
```suggestion
def test_kubernetes_task_config_enforces_command_requirements():
```
:p

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -392,36 +393,50 @@ def _pending_event_processing_loop(self) -> None:
...
-            )
+            # XXX: we were initially planning on using the name from KubernetesTaskConfig here,
+            # but its too easy to go over the length limit for container names (63 characters),
+            # so we're just hardcoding something for now since container names aren't used for
+            # anything at the moment
+            containers = [self._create_container_definition("main", task_config)]
+
+            for name, nested_config in task_config.extra_containers.items():
+                containers.append(self._create_container_definition(
+                    get_sanitised_kubernetes_name(name),
```
nemacysts, line 436:
```suggestion
                    get_sanitised_kubernetes_name(name, length_limit=63),
```

https://github.com/Yelp/task_processing/blob/850c64cc6346237ccffbe622aceb0f469ff582cd/task_processing/plugins/kubernetes/utils.py#L105 has a default of no length limit (could also replace this 63 with a constant)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: yea, afaict from reading the docs: ports in a pod need to be unique

nemacysts (commented):
should be good once we make sure folks aren't creating a task config with port collisions

nemacysts (approved):

PR #195, opened 2022-07-22
Adding an argument user_agent to Kubeclient

sidsatyelp:
https://jira.yelpcorp.com/browse/TRON-1743
set custom user agent for k8s client usage from tron/task_proc

Adding an argument user_agent to class KubeClient so that we can specify this from the caller.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -59,7 +59,7 @@ class KubernetesPodExecutor(TaskExecutor):
         task_configs: Optional[Collection[KubernetesTaskConfig]] = [],
     ) -> None:
         if not version:
-            version = "Unknown_Version"
+            version = "unknown_task_processing"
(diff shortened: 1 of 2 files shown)


EvanKrall (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -54,10 +54,14 @@ class KubernetesPodExecutor(TaskExecutor):
     def __init__(
         self,
         namespace: str,
+        version: Optional[str] = None,
         kubeconfig_path: Optional[str] = None,
         task_configs: Optional[Collection[KubernetesTaskConfig]] = [],
     ) -> None:
-        self.kube_client = KubeClient(kubeconfig_path=kubeconfig_path)
+        if not version:
+            version = "Unknown_Version"
```
nemacysts, line 62: thoughts on setting this to something like `unknown_task_processing`?

nemacysts (approved):

PR #197, opened 2022-08-11
Make it possible to receive all pod modified events

tzhu-yelp:
Some consumer is interested in a particular container's change in state after #194. These are sent as pod modified events since the other containers are still running. Such events are currently ignored since they don't result in a pod state transition, but if there are multiple containers, then it stands to reason for individual container state transitions to be something a caller could want to find out about.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -60,12 +60,20 @@ class KubernetesPodExecutor(TaskExecutor):
         version: Optional[str] = None,
         kubeconfig_path: Optional[str] = None,
         task_configs: Optional[Collection[KubernetesTaskConfig]] = [],
+        emit_events_without_state_transitions: bool = False,
+
     ) -> None:
         if not version:
             version = "unknown_task_processing"
         user_agent = f"{namespace}/v{version}"
         self.kube_client = KubeClient(kubeconfig_path=kubeconfig_path, user_agent=user_agent)
         self.namespace = namespace
+
+        # Pod modified events that did not result in a pod state transition are usually not
+        # forwarded, but some consumers are interested in container state changes. This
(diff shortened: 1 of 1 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -331,10 +339,22 @@ def __update_modified_pod(self, pod: V1Pod, event: Optional[PodEvent]) -> None:
...
-        )
+        if self.emit_events_without_state_transitions:
+            self.event_queue.put(
+                task_event(
+                    task_id=pod_name,
+                    terminal=False,
+                    timestamp=time.time(),
+                    raw=raw_event,
+                    task_config=task_metadata.task_config,
+                    platform_type="running",
```
nemacysts, line 350: one other thing we could probably do in the future (if it'd be helpful) is expand the types of task_events we enqueue (or add additional data (or take over `raw` and put custom data in there))

but this is probably good enough for a while

nemacysts (approved):

anjaberens (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nattofriends: what other events might be of interest?

tzhu-yelp: what other events might be of interest?

nemacysts: I was mostly just thinking we could make nice custom events for stuff like OOMKills like we had in Mesos (and maybe make a custom event for the main container exiting (but I guess that really bakes in the current hardcoded `main` container name))

fwiw, i don't think we should do this now: I was really just thinking out loud

nemacysts (approved):

PR #199, opened 2022-08-17
Okay to pass null disk requirements

tzhu-yelp:
Explicitly pass None to this field, receive no limit in Kubernetes

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -442,19 +442,21 @@ class KubernetesPodExecutor(TaskExecutor):
                 privileged=task_config.privileged,
             )
 
+            limits = {
+                "cpu": task_config.cpus,
+                "memory": f"{task_config.memory}Mi",
+            }
+
+            if task_config.disk is not None:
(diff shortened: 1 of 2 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -442,19 +442,21 @@ def _create_container_definition(
                 privileged=task_config.privileged,
             )
 
+            limits = {
+                "cpu": task_config.cpus,
+                "memory": f"{task_config.memory}Mi",
+            }
+
+            if task_config.disk is not None:
```
nemacysts, line 450: i'm a little iffy on this (especially if CI will be oncall for the pool that y'all end up using here) - having these disk limits helps us ensure that one workload doesn't bring down an entire box (or set of boxes) with high disk usage

is the motivation behind this change to help figure out what the actual usage is or will this be a permanent change for this workload?

tzhu-yelp: In this PR, I was thinking it would be a permanent change for this workload. I think there are some difficulties estimating the actual usage. For `main`, we could take the size of the passed artifact and multiply by some fixed number and that could work out reasonably. But I'm also concerned about dind for which the size of all images it will want to download isn't really known.

PR #201, opened 2023-03-15
Upgrades venv and tests to python 3.8

bobokun:
The reason for adding `@settings(suppress_health_check=(HealthCheck.function_scoped_fixture,))` was due to the warning below.

```
E       hypothesis.errors.FailedHealthCheck: Function-scoped fixture 'persister' used by 'tests/unit/plugins/persistence/dynamo_persistence_test.py::test_replaces_decimals_list'
E       
E       Function-scoped fixtures are not reset between examples generated by
E       `@given(...)`, which is often surprising and can cause subtle test bugs.
E       
E       If you were expecting the fixture to run separately for each generated example,
E       then unfortunately you will need to find a different way to achieve your goal
E       (e.g. using a similar context manager instead of a fixture).
E       
E       If you are confident that your test will work correctly even though the
E       fixture is not reset between generated examples, you can suppress this health
E       check to assure Hypothesis that you understand what you are doing.
E       
E       See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.function_scoped_fixture to the suppress_health_check settings for this test.

.tox/py38/lib/python3.8/site-packages/hypothesis/internal/healthcheck.py:27: FailedHealthCheck
```

Based on reading the tests and what they do I would assume that it's fine to re-use this fixture in multiple tests and suppress the warning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,48 +1,33 @@
+default_language_version:
+    python: python3
 repos:
-    - repo: https://github.com/pre-commit/pre-commit-hooks.git
-      rev: v2.0.0
-      hooks:
-          - id: trailing-whitespace
-            language_version: python3.8
-          - id: end-of-file-fixer
-            language_version: python3.8
-          - id: check-json
-            files: \.(jshintrc|json)$
-            language_version: python3.8
-          - id: check-yaml
-            language_version: python3.8
-          - id: debug-statements
-            language_version: python3.8
-          - id: name-tests-test
-            files: ^tests/.+\.py$
-            language_version: python3.8
-          - id: flake8
-            language_version: python3.8
-          - id: requirements-txt-fixer
-            language_version: python3.8
-          - id: check-added-large-files
-            language_version: python3.8
-          - id: check-byte-order-marker
-            language_version: python3.8
-          - id: fix-encoding-pragma
-            language_version: python3.8
-            args: [--remove]
-    - repo: https://github.com/asottile/reorder_python_imports.git
-      rev: v1.3.3
-      hooks:
-          - id: reorder-python-imports
-            language_version: python3.8
-            args:
-                [
-                    --remove-import,
-                    "from __future__ import absolute_import",
-                    --remove-import,
-                    "from __future__ import print_function",
-                    --remove-import,
-                    "from __future__ import unicode_literals",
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -1,44 +1,48 @@
...
--   repo: https://github.com/pre-commit/mirrors-autopep8
-    rev: v1.4.2
-    hooks:
-    -   id: autopep8
-        language_version: python3.6
+    - repo: https://github.com/pre-commit/pre-commit-hooks.git
+      rev: v2.0.0
+      hooks:
+          - id: trailing-whitespace
+            language_version: python3.8
```
nemacysts, line 6: we can make this easier for the next upgrade by using the `default_language_version` feature of pre-commit (and this also has the side-effect of cleaning up/minimizing the  config)

nemacysts (approved):
cleaning up the pre-commit config further would be nice for future-us, but that's not blocking

.pre-commit-config.yaml
bobokun: `default_language_version ` was giving an error but `python3` works

examples/cluster/playground/Dockerfile
```diff
@@ -1,27 +1,33 @@
...
-    wget http://mirrors.kernel.org/ubuntu/pool/universe/d/dh-virtualenv/dh-virtualenv_1.0-1_all.deb && \
-    gdebi -n dh-virtualenv*.deb && \
-    rm dh-virtualenv_*.deb
+ARG PIP_INDEX_URL
+ARG NPM_CONFIG_REGISTRY
+ENV PIP_INDEX_URL=${PIP_INDEX_URL:-https://pypi.python.org/simple}
+ENV NPM_CONFIG_REGISTRY=${NPM_CONFIG_REGISTRY:-https://npm.yelpcorp.com}
+
+# Get yarn
+# I'd use ubuntu yarn (yarnpkg) but https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1019291
```
nemacysts, line 25: is this still an issue? the linked bug says this should be fixed (which means hopefully we can also get rid of this workaround in tron)

nemacysts (approved):

PR #202, opened 2023-03-15
adds long_description

bobokun:
This adds long_description to PyPI.
https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/


setup.py
```diff
@@ -25,6 +29,8 @@
     author='Task Processing',
     author_email='team-taskproc@yelp.com',
     description='Framework for task processing executors and configuration',
+    long_description=long_description,
```
nemacysts, line 32: imo we could copy-paste the normal description here, but this seems cool too :)

nemacysts (approved):

PR #203, opened 2023-03-16
Adds reconciliation to pod tasks - multiprocessing

bobokun:
Adds a way to reconcile task metadata with Kubernetes task config
Todo: Add additional tests

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,5 +8,6 @@ pre-commit
 pytest
 pytest-cov
 pytest-mock
+pytest_bdd
 types-pyyaml
 types-requests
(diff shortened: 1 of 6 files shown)


nemacysts (commented):
we discovered the source of our issue: python added some locking/caching to some logger internals, and the way that tron (and the k8s client) work means that there is a logger leak that eventually causes task_processing to slow down

that said, it might still be nice to have a reconciliation loop in case something does happen to the watch loop (and i think it'd be helpful to start doing a more bulk-reconcile like we're starting to do here where we don't calling get_pods() N times - i'm pretty sure this is one of the causes of slow-ish restarts on some of our trons)

i think the only thing we'd maybe want to change here is moving away from `sleep(REFRESH_EXECUTOR_STATE_PROCESS_GRACE)` to having some sort of flag somewhere for when we're in a reconcilable state so that we don't attempt to double-reconcile in cases like a slow tron restart

nemacysts:
closing this for now until we come back to bringing a native reconcile loop into taskproc

PR #204, opened 2023-03-23
Support stdin and tty flags to support launching pods with commands like /bin/bash

danielpops:
## Problem
If you try to launch a kubernetes pod with command=`/bin/bash` it will exit immediately. I am working out how to use `paasta local-run` to run an interactive pod on the cluster, and this is a first step in that direction.

## Solution
Plumb through the `stdin`, `stdin_once`, and `tty` flags for the [V1Container](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1Container.md)


nemacysts (approved):
making local-run work in this way has been on my personal back-burner for a while, ty for picking this up!

PR #206, opened 2023-03-28
Adds reconciliation to pod tasks - threading

bobokun:
This is a threading approach to adding reconciliation to pod tasks.

Context: We're seeing Kubernetes events being missed by task_processing after upgrading Tron to Python 3.8. We're have not identified the root cause and are unable to reproduce this in our stage environment. However by performing reconciliation on the Kubernetes events we can be sure that events that are missed will be able to reconcile to its true state.

This is a similar approach to #203, but instead of multiprocessing we are using threading.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,8 +6,8 @@ mock
 mypy
 pre-commit
 pytest
+pytest-bdd
 pytest-cov
 pytest-mock
-pytest_bdd
 types-pyyaml
 types-requests
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,5 +8,6 @@ pre-commit
 pytest
 pytest-cov
 pytest-mock
+pytest_bdd
 types-pyyaml
 types-requests
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -436,7 +436,7 @@ class KubernetesPodExecutor(TaskExecutor):
         logger.info(f'Waiting {REFRESH_EXECUTOR_STATE_THREAD_GRACE}s before doing work')
         sleep(REFRESH_EXECUTOR_STATE_THREAD_GRACE)
         logger.debug("Starting Pod task config reconciliation.")
-        while not self.stopping:
+        while not self.is_stopping:
             try:
                 pods = self.kube_client.get_pods(namespace=self.namespace)
             except Exception:
@@ -675,3 +675,11 @@ class KubernetesPodExecutor(TaskExecutor):
 
     def get_event_queue(self) -> "Queue[Event]":
         return self.event_queue
+
+    @property
+    def is_stopping(self):
+        return self.stopping
+
+    @is_stopping.setter
+    def is_stopping(self, val):
+        self.stopping = val
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -187,21 +187,20 @@ class KubernetesPodExecutor(TaskExecutor):
                         "Exception encountered while watching Pod events - restarting watch!")
         logger.debug("Exiting Pod event watcher - stop requested.")
 
-    def _filter_task_configs_in_pods(
-            self,
-            pods: List[V1Pod]
+    def _group_pod_task_configs(
+        self,
+        pods: List[V1Pod]
     ) -> List[Tuple[KubernetesTaskConfig, V1Pod]]:
         """
         Called during reconciliation task loop in order to filter the task_configs/pods
         that are in task_metadata.
         """
-        v1pods = {pod.metadata.name: pod for pod in pods}
+        pods_seen = {pod.metadata.name: pod for pod in pods}
         task_config_pods = []
 
         for pod_name, task_metadata in self.task_metadata.items():
-            if pod_name in v1pods:
-                task_config_pod = (task_metadata.task_config, v1pods[pod_name])
-                task_config_pods.append(task_config_pod)
+            task_config_pod = (task_metadata.task_config, pods_seen.get(pod_name))
+            task_config_pods.append(task_config_pod)
 
         return task_config_pods
 
@@ -214,7 +213,7 @@ class KubernetesPodExecutor(TaskExecutor):
         that are mismatched in task_metadata.
         """
         task_configs_pods_to_reconcile = []
-        phase_task_state_map = {
+        phase_to_task_state = {
             "Succeeded": KubernetesTaskState.TASK_FINISHED,
             "Failed": KubernetesTaskState.TASK_FAILED,
             "Running": KubernetesTaskState.TASK_RUNNING,
@@ -222,24 +221,23 @@ class KubernetesPodExecutor(TaskExecutor):
             "Unknown": KubernetesTaskState.TASK_LOST
         }
 
-        if task_configs_pods:
-            for task_config, pod in task_configs_pods:
-                pod_name = pod.metadata.name
-                pod_phase = pod.status.phase
-                task_metadata = self.task_metadata[pod_name]
-                task_metadata_state = task_metadata.task_state
-                if pod_phase not in phase_task_state_map:
-                    logger.debug(
-                        f"Got a MODIFIED event for {pod_name} for unhandled phase: "
-                        f"{pod_phase} - ignoring."
-                    )
-                if phase_task_state_map[pod_phase] is not task_metadata_state:
-                    logger.debug(
-                        f"Mismatched event found for {pod_name} during reconciliation."
-                        f"pod_phase: {pod_phase} vs task_metadata_state: {task_metadata_state}"
-                        f"task_metadata{task_metadata}"
-                    )
-                    task_configs_pods_to_reconcile.append((task_config, pod))
+        for task_config, pod in task_configs_pods:
+            pod_name = pod.metadata.name
+            pod_phase = pod.status.phase
+            task_metadata = self.task_metadata[pod_name]
+            task_state = task_metadata.task_state
+            if pod_phase not in phase_to_task_state:
+                logger.debug(
+                    f"Got a MODIFIED event for {pod_name} for unhandled phase: "
+                    f"{pod_phase} - ignoring."
+                )
+            if phase_to_task_state[pod_phase] is not task_state:
+                logger.debug(
+                    f"Mismatched event found for {pod_name} during reconciliation. "
+                    f"pod_phase: {pod_phase} - task_state: {task_state} - "
+                    f"task_metadata: {task_metadata}"
+                )
+                task_configs_pods_to_reconcile.append((task_config, pod))
 
         return task_configs_pods_to_reconcile
 
@@ -514,13 +512,12 @@ class KubernetesPodExecutor(TaskExecutor):
                 # namespace - we'll now filter these by the tasks down to only
                 # those we know about so that we only reconcile the state for what
                 # we actually need and not any other cruft that may exist.
-                task_configs_pods = self._filter_task_configs_in_pods(pods)
+                task_configs_pods = self._group_pod_task_configs(pods)
                 # Filter for pods with mismatched states between K8s and task_metadata
                 task_configs_pods_to_reconcile = self._filter_task_configs_pods_to_reconcile(
                     task_configs_pods)
-                if task_configs_pods_to_reconcile:
-                    for task_config, pod in task_configs_pods_to_reconcile:
-                        self.reconcile(task_config, pod)
+                for task_config, pod in task_configs_pods_to_reconcile:
+                    self.reconcile(task_config, pod)
             logger.info(f'Sleeping for {self.refresh_reconciliation_thread_interval}s')
             sleep(self.refresh_reconciliation_thread_interval)
         logger.debug("Exiting Pod task config reconciliation - stop requested.")
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -187,6 +187,21 @@ class KubernetesPodExecutor(TaskExecutor):
                         "Exception encountered while watching Pod events - restarting watch!")
         logger.debug("Exiting Pod event watcher - stop requested.")
 
+    def _filter_task_configs_in_pods(
+            self,
+            pods: List[V1Pod]
+    ) -> List[Tuple[KubernetesTaskConfig, V1Pod]]:
+        """
+        Called during reconciliation task loop in order to filter the task_configs/pods
+        that are in task_metadata.
+        """
+        task_config_pods = []
+        for pod in pods:
+            if pod.metadata.name in self.task_metadata:
+                task_config_pod = (self.task_metadata[pod.metadata.name].task_config, pod)
+                task_config_pods.append(task_config_pod)
+        return task_config_pods
+
     def __handle_deleted_pod_event(self, event: PodEvent) -> None:
         pod = event["object"]
         pod_name = pod.metadata.name
@@ -443,7 +458,7 @@ class KubernetesPodExecutor(TaskExecutor):
         logger.info(f'Waiting {self.refresh_reconciliation_thread_grace}s before doing work')
         sleep(self.refresh_reconciliation_thread_grace)
         logger.debug("Starting Pod task config reconciliation.")
-        while not self.is_stopping:
+        while not self.stopping:
             try:
                 # fetch all pods in the target namespace in one request so that we
                 # don't block on making N serial requests to the Kubernetes API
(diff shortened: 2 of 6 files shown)


task_processing/utils.py
```diff
@@ -39,3 +39,13 @@ def get_cluster_master_by_proxy(
...
+
+
+def strtobool(s: str) -> bool:
+    truevals = ("yes", "y", "on", "true", "t", "1")
+    falsevals = ("no", "n", "off", "false", "f", "0")
+    if s.lower() in truevals:
+        return True
+    if s.lower() in falsevals:
+        return False
+    raise ValueError(f"invalid truth value: {s}")
```
nemacysts, line 51: we can probably get rid of this once we move the toggle for reconciliation to a plugin param :)

tests/unit/plugins/kubernetes/kube_client_test.py
```diff
@@ -111,3 +111,22 @@ def test_KubeClient_get_pod():
...
+    ) as mock_kube_client, mock.patch.dict(
+        os.environ, {"KUBECONFIG": "/another/kube/config.conf"}
+    ):
+        mock_config_path = "/OVERRIDE.conf"
+        mock_kube_client.CoreV1Api().list_namespaced_pod.return_value = mock.Mock()
+        client = KubeClient(kubeconfig_path=mock_config_path)
+        client.get_pods(namespace='ns', attempts=1)
+    mock_kube_client.CoreV1Api().list_namespaced_pod.assert_called_once_with(
+        namespace='ns'
+    )
```
nemacysts, line 132: might be good to also add some tests for the exception cases (the happy path test for this function and `get_pod()` on their own aren't particularly high-value without the failure case tests)

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -318,6 +318,55 @@ def test_pending_event_processing_loop_processes_remaining_events_after_stop(k8s
...
+        mock_pod.status.host_ip = '1.2.3.4'
+        mock_pod.spec.node_name = 'kubenode'
+        mock_pods.append(mock_pod)
+        task_metadata[taskconf.pod_name] = KubernetesTaskMetadata(
+            task_config=taskconf,
+            task_state=KubernetesTaskState.TASK_RUNNING,
+            task_state_history=v(),
+        )
+    # Keep only 3 task_configs in task_metadata
+    task_metadata.popitem()
```
nemacysts, line 347: can we generate just the 3 tasks to minimize the logic in this test?

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
```diff
@@ -318,6 +318,55 @@ def test_pending_event_processing_loop_processes_remaining_events_after_stop(k8s
...
+    ) as mock_kube_client, mock.patch(
+        "task_processing.plugins.kubernetes.kubernetes_pod_executor.KubernetesPodExecutor.is_stopping",  # noqa
+        new_callable=mock.PropertyMock,
+        side_effect=[False, True]
+    ):
+        mock_kube_client.get_pods.return_value = mock_pods
+        k8s_executor._reconcile_task_loop()
+
+    mock_reconcile_task.assert_called()
+    assert mock_reconcile_task.call_count == 3
```
nemacysts, line 367: should we add an assert that the reconcile loop either changed/didn't change anything?

nemacysts (commented):
i'm a little worried about performance in some of our bigger tron clusters, but we can definitely give this a try (and/or maybe load-test this with jolt first)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -49,6 +51,12 @@
     "Succeeded",
     "Unknown",
 }
+REFRESH_EXECUTOR_STATE_THREAD_GRACE: int = int(
+    os.getenv("TASK_PROCESSING_REFRESH_EXECUTOR_STATE_THREAD_GRACE", 300))
+REFRESH_EXECUTOR_STATE_THREAD_INTERVAL: int = int(
+    os.getenv("TASK_PROCESSING_REFRESH_EXECUTOR_STATE_THREAD_INTERVAL", 120))
+ENABLE_RECONCILIATION: bool = strtobool(
+    os.getenv("TASK_PROCESSING_RECONCILIATION_ENABLE", "True"))
```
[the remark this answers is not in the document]
bobokun: Just do be clear do you mean adding it as a parameter to the KubernetesPodExecutor class in the `__init__` section?

requirements-dev.txt
```diff
@@ -8,5 +8,6 @@ pre-commit
 pytest
 pytest-cov
 pytest-mock
+pytest_bdd
```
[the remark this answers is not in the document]
nemacysts: gotcha - in that case: 
```suggestion
pytest-bdd
```
should be the requirement name

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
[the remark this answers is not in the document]
nemacysts: yup - an example of a recent addition like this is the `emit_events_without_state_transitions` toggle

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -635,3 +681,11 @@ def stop(self) -> None:
...
     def get_event_queue(self) -> "Queue[Event]":
         return self.event_queue
+
+    @property
+    def is_stopping(self):
+        return self.stopping
+
+    @is_stopping.setter
+    def is_stopping(self, val):
+        self.stopping = val
```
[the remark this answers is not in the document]
bobokun: I was having issues with writing a test for `_reconcile_task_loop` as it was entering an infinite loop while waiting for `self.stopping` to turn true during the test case. In order to avoid this I mock the is_stopping function here in order to simulate only 1 loop and exit out after the second loop. https://github.com/Yelp/task_processing/blob/766c7e058d1cde9401f6bab4d58d4ee582e1ead0/tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py#L359-L361

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
bobokun: the reconcile function here is actually mocked so it won't actually modify the events. This test here is simply to ensure that we are calling the reconcile function the right number of times based on what we see mock_pods/task_metadata. 

The tests for actually reconciling is handled separately in other tests.

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
bobokun: This is actually intentional, since we aren't testing the actual reconciliation function I wanted to see if we had 4 V1Pods from kubernetes but only 3 in task_metadata it would reconcile those 3 tasks and ignore the 4th pod

nemacysts: ah gotcha - might be good to explicitly document that in this comment :)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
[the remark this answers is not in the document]
bobokun: great idea

tests/unit/plugins/kubernetes/kubernetes_pod_executor_test.py
bobokun: added some additional comments :)

nemacysts (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -420,6 +451,37 @@ def _pending_event_processing_loop(self) -> None:
...
+        while not self.stopping:
+            try:
+                # fetch all pods in the target namespace in one request so that we
+                # don't block on making N serial requests to the Kubernetes API
+                pods = self.kube_client.get_pods(namespace=self.namespace)
+            except Exception:
+                logger.exception(
+                    f"Hit an exception attempting to fetch pods in namespace {self.namespace}")
+                pods = None
+                continue
```
jfongatyelp, line 470: Because we're `continue`ing here before the sleep interval, if something is wrong with either the api or tron's connection to the api, this seems like it could get us into a constant infinite loop of attempting to fetch the pods w/o pauses in between, which could overwhelm the k8s api (or even this reconciliation thread?). 

We might want to avoid that and add some backoff here if we hit an exception.

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -420,6 +451,37 @@ def _pending_event_processing_loop(self) -> None:
...
+
+            if pods is not None:
+                # we've previously bulk-fetched all the pods running in the target
+                # namespace - we'll now filter these by the tasks down to only
+                # those we know about so that we only reconcile the state for what
+                # we actually need and not any other cruft that may exist.
+                task_configs_pods = self._filter_task_configs_in_pods(pods)
+                if task_configs_pods:
+                    for task_metadata, pod in task_configs_pods:
```
jfongatyelp, line 479: Since _filter_task_configs_in_pods will only return tasks with pods found from k8s api, this means we'll never actually reconcile tasks we know about but *don't* have pods.  That seems like we might still have the edge case where we can miss pods going away and never being reconciled by tron/setting them to `LOST`.  

Perhaps we should iterate over all task_metadata instead of only those with discovered pods.

jfongatyelp (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
bobokun: hm, I'm not sure why I added this continue here. I don't think we actually need it

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
bobokun: good catch! I'll update by iterating over task_metadata instead :)

nemacysts: jfong++ - i totally tunnel-visioned/brain-fog'd over this case!

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -171,6 +187,56 @@ def _pod_event_watch_loop(self) -> None:
...
+        """
+        v1pods = {pod.metadata.name: pod for pod in pods}
+        task_config_pods = []
+
+        for pod_name, task_metadata in self.task_metadata.items():
+            if pod_name in v1pods:
+                task_config_pod = (task_metadata.task_config, v1pods[pod_name])
+                task_config_pods.append(task_config_pod)
+
+        return task_config_pods
```
nemacysts, line 206: hmm, I think that we're still not covering the case that jfong mentioned: let's say tron has some sort of outage at the same time as some sort of k8s outage (or human k8s cleanup) - when tron/task_proc restart it's possible that we have some state stored for a previously running pod and k8s no longer has any events/pod metadata.

in this case: we won't return that task_config -> we won't see it in `_filter_task_configs_pods_to_reconcile` -> we'll have an "orphaned" task_config

i think perhaps what we want to do here is maybe something like
```suggestion
        pods_seen = {pod.metadata.name: pod for pod in pods}
        task_config_pods = []

        for pod_name, task_metadata in self.task_metadata.items():
            task_config_pod = (task_metadata.task_config, pods_seen.get(pod_name))
            task_config_pods.append(task_config_pod)

        return task_config_pods
```
and then treat having a `None` for the `V1Pod` entry in the tuple as a LOST event

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -171,6 +187,56 @@ def _pod_event_watch_loop(self) -> None:
...
+        task_configs_pods_to_reconcile = []
+        phase_task_state_map = {
+            "Succeeded": KubernetesTaskState.TASK_FINISHED,
+            "Failed": KubernetesTaskState.TASK_FAILED,
+            "Running": KubernetesTaskState.TASK_RUNNING,
+            "Pending": KubernetesTaskState.TASK_PENDING,
+            "Unknown": KubernetesTaskState.TASK_LOST
+        }
+
+        if task_configs_pods:
```
nemacysts, line 225: we can get rid of this conditional and unident the contents here: attempting to iterate over an empty list is a noop:
```
>>> for a,b in []:
...     print(f"{a} - {b}")
...
>>>
```

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -171,6 +187,56 @@ def _pod_event_watch_loop(self) -> None:
...
+    def _filter_task_configs_pods_to_reconcile(
+        self,
+        task_configs_pods: List[Tuple[KubernetesTaskConfig, V1Pod]]
+    ) -> List[Tuple[KubernetesTaskConfig, V1Pod]]:
+        """
+        Called during reconciliation task loop in order to filter the task_configs/pods
+        that are mismatched in task_metadata.
+        """
+        task_configs_pods_to_reconcile = []
+        phase_task_state_map = {
```
nemacysts, line 217: thoughts on `phase_to_task_state` as a name for this?

(feel free to ignore)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -171,6 +187,56 @@ def _pod_event_watch_loop(self) -> None:
...
+            "Running": KubernetesTaskState.TASK_RUNNING,
+            "Pending": KubernetesTaskState.TASK_PENDING,
+            "Unknown": KubernetesTaskState.TASK_LOST
+        }
+
+        if task_configs_pods:
+            for task_config, pod in task_configs_pods:
+                pod_name = pod.metadata.name
+                pod_phase = pod.status.phase
+                task_metadata_state = self.task_metadata[pod_name].task_state
```
nemacysts: thoughts on `task_state` as a name for this?
(feel free to ignore)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -420,6 +486,39 @@ def _pending_event_processing_loop(self) -> None:
...
+                logger.exception(
+                    f"Hit an exception attempting to fetch pods in namespace {self.namespace}")
+                pods = None
+
+            if pods is not None:
+                # we've previously bulk-fetched all the pods running in the target
+                # namespace - we'll now filter these by the tasks down to only
+                # those we know about so that we only reconcile the state for what
+                # we actually need and not any other cruft that may exist.
+                task_configs_pods = self._filter_task_configs_in_pods(pods)
```
nemacysts, line 517: thoughts on renaming this function to `_group_pod_task_configs()` so that its name matches what it should be doing more closely?

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -420,6 +486,39 @@ def _pending_event_processing_loop(self) -> None:
...
+                # Filter for pods with mismatched states between K8s and task_metadata
+                task_configs_pods_to_reconcile = self._filter_task_configs_pods_to_reconcile(
+                    task_configs_pods)
+                if task_configs_pods_to_reconcile:
+                    for task_config, pod in task_configs_pods_to_reconcile:
```
nemacysts, line 522: we can probably do the same as above and rely on iterating over an empty list being a noop to reduce a level of indentation :)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -171,6 +187,62 @@ def _pod_event_watch_loop(self) -> None:
...
+                    logger.debug(
+                        f"Got a MODIFIED event for {pod_name} for unhandled phase: "
+                        f"{pod_phase} - ignoring."
+                    )
+                if phase_task_state_map[pod_phase] is not task_metadata_state:
+                    logger.debug(
+                        f"Mismatched event found for {pod_name} during reconciliation."
+                        f"pod_phase: {pod_phase} vs task_metadata_state: {task_metadata_state}"
+                        f"task_metadata{task_metadata}"
+                    )
```
nemacysts, line 241: i think the formatting here might be a little off?
```suggestion
                    logger.debug(
                        f"Mismatched event found for {pod_name} during reconciliation. "
                        f"pod_phase: {pod_phase} - task_metadata_state: {task_metadata_state} - "
                        f"task_metadata: {task_metadata}"
                    )
```
should fix the spacing (i also took the liberty of adding a consistent separator :))

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -171,6 +187,63 @@ def _pod_event_watch_loop(self) -> None:
...
+        phase_to_task_state = {
+            "Succeeded": KubernetesTaskState.TASK_FINISHED,
+            "Failed": KubernetesTaskState.TASK_FAILED,
+            "Running": KubernetesTaskState.TASK_RUNNING,
+            "Pending": KubernetesTaskState.TASK_PENDING,
+            "Unknown": KubernetesTaskState.TASK_LOST
+        }
+
+        for task_config, pod in task_configs_pods:
+            # If the pod is not returned then we must set the task_metadata to lost state
```
nemacysts, line 225:
```suggestion
            # If the pod is not returned then we must set the task_metadata to lost state - if k8s no longer knows about this pod/task, then there's no way for us to figure out it's final state in an automated fashion
```

nemacysts (approved):
i have a minor comment suggestion, but feel free to fix'n'ship or drop the suggestion and ship :)

nemacysts:
we discovered the source of our issue: python added some locking/caching to some logger internals, and the way that tron (and the k8s client) work means that there is a logger leak that eventually causes task_processing to slow down

PR #207, opened 2023-04-13
Add support for overriding resource requests

anjaberens:
Resource requests are currently not explicitly set - this results in k8s setting
the memory/cpu/disk requests to the same value as the limit.

For the most part, this is desired behaviour at Yelp as the main consumer of this
library is Tron and we don't expect our batches to need to "burst" any of these
resources. Additionally, we generally want to signal to k8s that our batches shouldn't
be killed if there are other (more interrupt-tolerant) workloads on an overloaded
node by defaulting the QoS class for any task_proc-created Pods to `Guaranteed`
precisely by setting `requests = limits`.

However, one of our other large task_proc workloads (Jolt) would like to have 
control over the requests as their workload runs in an isolated pool and cannot (at
least at the moment) be easily right-sized for optimal bin-packing/resource usage.

But, given that Jolt runs in its own pool, we can be a little cheeky and set lower
requests to pack Pods a little tighter since Jolt *is* quite interrupt-tolerant and
as such is resilient to its Pods getting killed due to node resource pressure (which
should be pretty rare in any case given the current usage patterns).

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -4,7 +4,9 @@ import threading
 import time
 from queue import Queue
 from typing import Collection
+from typing import Dict
 from typing import Optional
+from typing import Union
 
 from kubernetes import watch
 from kubernetes.client import V1Affinity
@@ -442,19 +444,38 @@ class KubernetesPodExecutor(TaskExecutor):
                 privileged=task_config.privileged,
             )
 
+        # Set any optional requests
+        requests: Dict[str, Union[str, float]] = {}
+        if 'request_cpus' in task_config and task_config.request_cpus:
+            requests['cpus'] = task_config.request_cpus
+        if 'request_memory' in task_config and task_config.request_memory:
+            requests['memory'] = f"{task_config.request_memory}Mi"
+        if 'request_disk' in task_config and task_config.request_disk:
+            requests['ephemeral-storage'] = f"{task_config.request_disk}Mi"
+
+        limits = {
+            "cpu": task_config.cpus,
+            "memory": f"{task_config.memory}Mi",
+            "ephemeral-storage": f"{task_config.disk}Mi",
+        }
+
+        if not requests:
+            resources = V1ResourceRequirements(
+                limits=limits,
+            )
+        else:
+            resources = V1ResourceRequirements(
+                limits=limits,
+                requests=requests,
+            )
+
         return V1Container(
             image=task_config.image,
             name=name,
             command=["/bin/sh", "-c"],
             args=[task_config.command],
             security_context=security_context,
-            resources=V1ResourceRequirements(
-                limits={
-                    "cpu": task_config.cpus,
-                    "memory": f"{task_config.memory}Mi",
-                    "ephemeral-storage": f"{task_config.disk}Mi",
-                }
-            ),
+            resources=resources,
             env=get_kubernetes_env_vars(
                 environment=task_config.environment,
                 secret_environment=task_config.secret_environment,
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -4,9 +4,7 @@ import threading
 import time
 from queue import Queue
 from typing import Collection
-from typing import Dict
 from typing import Optional
-from typing import Union
 
 from kubernetes import watch
 from kubernetes.client import V1Affinity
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -442,19 +444,38 @@ def _create_container_definition(
...
             )
 
+        # Set any optional requests
+        requests: Dict[str, Union[str, float]] = {}
+        if 'request_cpus' in task_config and task_config.request_cpus:
+            requests['cpus'] = task_config.request_cpus
+        if 'request_memory' in task_config and task_config.request_memory:
+            requests['memory'] = f"{task_config.request_memory}Mi"
+        if 'request_disk' in task_config and task_config.request_disk:
+            requests['ephemeral-storage'] = f"{task_config.request_disk}Mi"
```
nemacysts, line 454: imo, it'd be nice to keep the disk stuff as-is since I don't think we have much of a system reservation for disk (if we have any at all) - but if this is just for completeness, it should be fine to leave :)

nemacysts, line 454: if we set a default value of `None` for these, we should be able to avoid the `if FOO in task_config` bits here 


it's also maybe worth testing to see what happens if you feed k8s a null request as maybe we can build the `request` dict in the same way as the `limits` one - e.g.,
```python
requests = {
            "cpu": task_config.request_cpus if task_config.request_cpu else None,
            "memory": f"{task_config.request_memory}Mi"  if task_config.request_memory else None,
            "ephemeral-storage": f"{task_config.request_disk}Mi" if task_config.request_disk else None,
}
limits = {
            "cpu": task_config.cpus,
            "memory": f"{task_config.memory}Mi",
            "ephemeral-storage": f"{task_config.disk}Mi",
}
resources = V1ResourceRequirements(
                limits=limits,
                requests=requests,
)
```

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -304,6 +304,17 @@ def __invariant__(self) -> Tuple[Tuple[bool, str], ...]:
...
+
+    request_cpus = field(
+        type=float,
+        factory=float)
+    request_memory = field(
+        type=float,
+        factory=float)
+    request_disk = field(
+        type=float,
+        factory=float)
```
nemacysts, line 316: imo, `$RESOURCE_request` would read a tad better :)

that said, for my suggestion above about giving these a null initial value:
```suggestion
    request_cpus = field(
        type=(float, type(None)),
        initial=None,
    )
    request_memory = field(
        type=(float, type(None)),
        initial=None,
    )
    request_disk = field(
        type=(float, type(None)),
        initial=None,
    )
```

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
anjaberens: Its only for completeness. I can remove if you have strong feelings about this.

anjaberens: I did try to set the default to None, iirc it caused the type system to get mad and make me do Union(None | float) kind of thing, hence this evasion.

task_processing/plugins/kubernetes/task_config.py
anjaberens: Flipping the name totally makes sense, i mostly worded it as request_x because of how it is in the structure itself.

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
nemacysts: yea, I think without the `type=REAL_TYPE, type(None)` bit pyrsistent gets quite upset with None defaults :p

nemacysts: yea, I say we can circle back and add this later if we ever end up needing this - it makes me a tad nervous as-is tho (mostly 'cause I'm not sure if the kubelet will protect us by evicting things when the node is under disk pressure or if we'll just run the node out of disk and kill it :p)

anjaberens: I set this up and it works, main wrinkle is without the factory it gets upset when it gets a integer value (ie 768) for the fields. I'm not sure how much we care about this or should i go ahead and create a float_or_none factory?

nemacysts: that sounds pretty reasonable! i can definitely see folks passing in integer values for these

anjaberens: For the records - the testing this on jolt works. I will follow up next week with a factory for this.

nemacysts (approved):

task_processing/plugins/kubernetes/task_config.py
```diff
@@ -238,6 +238,12 @@ def _valid_service_account_name(
     )
 
 
+def _float_or_none(str: Optional[str]) -> Optional[float]:
+    if str:
+        return float(str)
+    return None
```
nemacysts, line 244: should this be something like:
```suggestion
def _float_or_none(val: Any) -> Optional[float]:
    return float(val) if val is not None else None
```

otherwise, the type hints aren't exactly correct (`_float_or_none(1)` is a valid call, but `1` is not a string.

thinking about this line also made me realize that I think we want an `invariant` set for the `*_request` fields as they should always be non-zero if set (I was going to say that the current factory would return `None` if passed in `0` or `0.0`, but then I realized that we don't really want 0-valued requests in the first place)

that said, I'm not quite sure what the ordering is here (i.e., is `factory()` called first or is `invariant()`?) so we'll probably need to do some testing here

anjaberens: seems to run the invariant after factory i believe (?)

nemacysts (approved):

PR #210, opened 2023-06-22
Only set explicit requests if a value is set 

nemacysts:
The previous code did not do what we expected and ended up creating
pods with a limit and no request.


jfongatyelp (approved):

bobokun (approved):
nice, lgtm! Thanks for adding all the different tests that cover all the possibilities

Molaire (approved):

PR #211, opened 2024-01-04
Reusing global Configuration() object in k8s client models instances

cuza:
Reusing global Configuration() object in k8s client models instances to avoid calling setLevel which will do a potentially expensive cache clear

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -16,6 +16,7 @@ from kubernetes.client import V1PodSecurityContext
 from kubernetes.client import V1PodSpec
 from kubernetes.client import V1ResourceRequirements
 from kubernetes.client import V1SecurityContext
+from kubernetes.client.configuration import Configuration
 from kubernetes.client.exceptions import ApiException
 from pyrsistent import pmap
 from pyrsistent import v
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -43,6 +44,7 @@
 from task_processing.plugins.kubernetes.utils import get_sanitised_kubernetes_name
 
 logger = logging.getLogger(__name__)
+k8s_client_configuration = Configuration.get_default_copy()
```
nemacysts, line 47: we should probably UPPER_CASE this to make it more obvious that this is a global variable

nemacysts: I also wonder if it's worth having this in a central location so we truly only ever make a single Configuration object (although, going down to 2 is still a massive improvement :p)

nemacysts (approved):
it'd be nice to UPPER_CASE the config variable :)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
cuza: I do like the idea of the central location, we could even use the same Configuration() object and even pass it on to the ApiClient()

nemacysts (approved):
@cuza think it's still worth trying to merge this? i'm a tad worried that folks will forget to add this config so we'd probably want to pair this with a pre-commit hook or something :/

nemacysts:
closing this for now since we went with patching the k8s clientlib instead

PR #212, opened 2024-01-19
Handling Added events

EmanElsaban:
This PR currently contains: 
- handling of Added events

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -205,12 +205,13 @@ class KubernetesPodExecutor(TaskExecutor):
         """Called during reconciliation and normal event handling"""
         pod_name = pod.metadata.name
         task_metadata = self.task_metadata[pod_name]
+        event_type = event["type"] if event else "null"
 
         raw_event = event["raw_object"] if event else None
 
         if pod.status.phase not in SUPPORTED_POD_MODIFIED_EVENT_PHASES:
             logger.debug(
-                f"Got a MODIFIED event for {pod_name} for unhandled phase: "
+                f"Got a {event_type} event for {pod_name} for unhandled phase: "
                 f"{pod.status.phase} - ignoring."
             )
             return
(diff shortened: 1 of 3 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -388,7 +471,7 @@ def _process_pod_event(self, event: PodEvent) -> None:
             elif event["type"] == "DELETED":
                 self.__handle_deleted_pod_event(event)
 
-            elif event["type"] == "MODIFIED":
+            elif event["type"] == "MODIFIED" or event["type"] == "ADDED":
```
nemacysts, line 453:
```suggestion
            elif event["type"] in {"MODIFIED", "ADDED"}:
```
would be a little more idiomatic/shorter, but the current code works as well :)

nemacysts (commented):
(we also chatted a bit internally)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -404,6 +404,10 @@ def _pending_event_processing_loop(self) -> None:
         event = None
         while not self.stopping or not self.pending_events.empty():
             try:
+                # we might see that their are gaps 0.5s because thats how long it will take it to see if there are stuff in the queue
+                # will give you whats in the queue or wait 0.5 sec to receive events, if no events are received then it will throw the empty exception
+                # and start again      
+                # I think below might be taking some time to get from a queue an event, should time these two separately
```
nemacysts, line 410: we can probably remove this since there's no more timers here :)

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -210,7 +210,7 @@ def __update_modified_pod(self, pod: V1Pod, event: Optional[PodEvent]) -> None:
 
         if pod.status.phase not in SUPPORTED_POD_MODIFIED_EVENT_PHASES:
             logger.debug(
-                f"Got a MODIFIED event for {pod_name} for unhandled phase: "
+                f"Got a {event['type']} event for {pod_name} for unhandled phase: "
```
nemacysts, line 213: i wonder if we can do something like:
```python
    def __update_modified_pod(self, pod: V1Pod, event: Optional[PodEvent]) -> None:
        ...
        event_type = event['type'] if event else "null"
        ...
```
to handle the null event case (from reconciling)

nemacysts (approved):

KaspariK (approved):

PR #213, opened 2024-04-26
A quick attempt at making taskproc handle migrations from one k8s clu…

EvanKrall:
…ster to another

This change has been tested in one of our test tron servers and is able to properly update and reconcile tasks if they are found in either the current kube cluster or one accessed via old_kubeconfig_paths.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -662,9 +662,6 @@ class KubernetesPodExecutor(TaskExecutor):
         This function will request that Kubernetes delete the named Pod and will return
         True if the Pod termination request was succesfully emitted or False otherwise.
         """
-        # NOTE: we're purposely not removing this task from `task_metadata` as we want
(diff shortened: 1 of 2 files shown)


task_processing/plugins/kubernetes/kubernetes_pod_executor.py
```diff
@@ -72,13 +72,20 @@ def __init__(
         kubeconfig_path: Optional[str] = None,
         task_configs: Optional[Collection[KubernetesTaskConfig]] = [],
         emit_events_without_state_transitions: bool = False,
+        old_kubeconfig_paths: Collection[str] = (),
```
nemacysts, line 75: might be worth adding a comment (or having a more generic name, e.g. watcher_kubeconfig_paths) in case this sticks around past our migration :p

nemacysts (approved):

task_processing/plugins/kubernetes/kubernetes_pod_executor.py
jfongatyelp: TY! I like that naming better

nemacysts (approved):

PR #215, opened 2024-05-10
Delete all Mesos-using code (+ misc. cleanups)

nemacysts:
1. Clean-up makefile (removed all dead targets, made all targets PHONY)
2. Remove all unused plugins (deleted code is code we don't need to maintain)
3. Regenerate docs (target was not phony, so it would never regenerate docs unless run with --always-make)
4. Clean-up README.md (removed Mesos references as well as references to deleted plugins and whatnot)
5. Cleaned up dead tests (this included deleting all the itests since they were all Mesos-related)


jfongatyelp (approved):
til of many of these plugins we're deleting 😅

bobokun (approved):
🎉

PR #218, opened 2024-06-27
Allow task_processing-launched pods to handle their own cleanup

nemacysts:
We have some usecases where we'd like a set of batches to run some cleanup code upon recieving a SIGTERM, so we'll need to remove our force-kill code and instead fallback to the default k8s behavior of giving Pods 30s after sending a SIGTERM before sending a SIGKILL to force-terminate :)

I will also follow this up with a Tron documentation change that documents this behavior.

### DIFF
--- a/task_processing/plugins/kubernetes/kube_client.py
+++ b/task_processing/plugins/kubernetes/kube_client.py
@@ -99,12 +99,17 @@ class KubeClient:
         while attempts:
             try:
                 logger.info(f"Attempting to terminate {pod_name}")
+                # NOTE: we used to force-kill pods by setting grace_period_seconds=0, but we later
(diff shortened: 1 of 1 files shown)


jfongatyelp (approved):
If we do this is it possible task_proc threads get blocked for 30s while just waiting for this deletion to finish? Or is python's async stuff smart enough to continue doing $other things in the meantime?

nemacysts:
> If we do this is it possible task_proc threads get blocked for 30s while just waiting for this deletion to finish? Or is python's async stuff smart enough to continue doing $other things in the meantime?

err, that's a good question - i think we'd probably want to make this an async request (which will require some testing since we don't currently use the client to make async requests anywhere)

PR #221, opened 2024-12-19
Only drop capabilities that are not added

nemacysts:
Just like https://github.com/Yelp/paasta/pull/3972 and https://github.com/Yelp/paasta/pull/3973, we need to ensure that
there are no duplicates between cap_add and cap_drop - otherwise,
the cap_drop entry will "win" and the duplicate capability will not be
added.


nemacysts:
cc @anjaberens - doubt y'all have run into this in Jolt, but just in case :)

jfongatyelp (approved):

PR #222, opened 2025-01-25
TRON-2333: Fix pod names ending in .--xxxx

jfongatyelp:
There was an edge case where sanitized pod names ending up with .<6 characters> would then get truncated to something like `.--<4 random chars>` due to our truncation logic in `get_sanitised_kubernetes_name`. 

However `.--aaaa` is not valid in a pod name as each section between `.`s must start with an alphanumeric character.

We saw errors such as the following when this case was hit:
```
{"reason":"FieldValueInvalid","message":"Invalid value: \"some-long-pod-name.--9b3a\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and
must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')","field":"metadata.name"}
```

## Solution
Before we try to add our truncation suffix, we simply strip any trailing `.` in our truncated name. This yields a pod name 1 character less than expected, but ensures pod name formats will continue to match previous expectations. 

I also did not strip any further characters as `.` was the most important one that would cause validation to fail.

## Verification
with the new test in place but using the old `get_sanitised_kubernetes_name`, the test fails as expected, with the non-RFC1123-compliant result:
```
E       AssertionError: assert 'hello--world.--cf28' == 'hello--world--cf28'
E
E         - hello--world--cf28
E         + hello--world.--cf28
E         ?             +
```

### DIFF
--- a/task_processing/plugins/kubernetes/utils.py
+++ b/task_processing/plugins/kubernetes/utils.py
@@ -153,8 +153,10 @@ def get_sanitised_kubernetes_name(
         # the part of the name that does fit in the limit in order to replace them with
         # -- (2 characters) and then a 4 character hash (which should help ensure uniqueness
         # after truncation).
+        truncated_name = name[0 : length_limit - 6]
         name = (
-            name[0 : length_limit - 6]
+            # Avoid splitting the suffix into a subdomain as it would not meet RFC1123 requirements
+            truncated_name.rstrip(".")
             + "--"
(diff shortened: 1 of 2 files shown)


nemacysts (approved):
my bad, i could have sworn i had shipped this earlier!

PR #223, opened 2025-03-25
Handle Watch expiration by resetting the resourceVersion

nemacysts:
We've seen an increase in cases where the Watch (for a reason I don't quite understand yet - some of the things I've been reading suggest that there's maybe something re: k8s API caches and whatnot exipring) starts erroring out with exceptions like:
```
kubernetes.client.exceptions.ApiException: (410)
Reason: Expired: too old resource version: 1642912881 (1644329743
```

While in the past this seems to have generally always fixed itself with the pre-existing restart code, in the aforementioned cases we've seen this loop get stuck in a restart-loop that it we either never recover from until a human intervenes OR it takes long enough to recover that we start losing events.

This diff ensures that when we restart the watch loop, we restart as if we were starting a fresh Watch - this does mean that we'll re-process existing events as the current behavior of a `Watch().stream()` includes returning the current state, but that's fine since while GH issues mention that this isn't necessarily ordered, our state machine/transition logic should ensure that out-of-order events don't actually hurt us - at worst, we'll simply waste time.

### DIFF
--- a/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
+++ b/task_processing/plugins/kubernetes/kubernetes_pod_executor.py
@@ -163,14 +163,12 @@ class KubernetesPodExecutor(TaskExecutor):
         self, kube_client: KubeClient, watch: kube_watch.Watch
     ) -> None:
         logger.debug(f"Starting watching Pod events for namespace={self.namespace}.")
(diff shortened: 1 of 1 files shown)


nemacysts:
I tested this on infrastage, but this HTTP410 issue doesn't seem to really happen in this cluster (and it's not particularly frequent enough in any dev clusters for me to be able to temporarily test this there)

jfongatyelp (approved):

